PR #1, opened 2021-02-26
Add domain bridge library

jacobperron:
This is an initial implementation that supports bridging arbitrary topics across ROS 2 domains.
The generic publisher and generic subscription are directly copied from rosbag2, but I'd like to switch to a common implementation once it is ready (see https://github.com/ros2/rclcpp/pull/1452).

I've also included a README with aspirational usage and a design doc.

This PR does not include some core features that I'd like to add in follow-up PRs:

- Correctly mapping QoS settings across the bridge
- YAML configuration


jacobperron:
cc @mallanmba

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,218 @@
...
+    this->executor_context_ = create_context_with_domain_id(1);
+    rclcpp::ExecutorOptions executor_options;
+    executor_options.context = this->executor_context_;
+    this->executor_ = std::make_unique<rclcpp::executors::MultiThreadedExecutor>(executor_options);
+  }
+
+  ~DomainBridgeImpl()
+  {
+    const std::string reason("Domain bridge shutdown");
+    this->executor_context_->shutdown(reason);
```
ivanpauno, line 56:
```suggestion
    this->executor_context_->shutdown("Domain bridge shutdown");
```

?

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,218 @@
...
+  rclcpp::Node::SharedPtr get_node_for_domain(size_t domain_id)
+  {
+    auto domain_id_node_pair = node_map_.find(domain_id);
+
+    // If we don't already have a node for the domain, create one
+    if (node_map_.end() == domain_id_node_pair) {
+      auto context = create_context_with_domain_id(domain_id);
+      auto node_options = create_node_options(context);
+      std::ostringstream oss;
+      oss << "domain_bridge_" << std::to_string(domain_id);
```
ivanpauno, line 86: There could be potentially more than one domain bridge.
I would use something like: `domain_bridge_{domain_bridge_id}_{domain_id}`

Where the "domain_bridge_id" is a random number representing the bridge (or anything that works a pseudo-unique id for the "bridge").

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,218 @@
...
+    const std::string & topic_name,
+    const rosidl_message_type_support_t & typesupport_handle)
+  {
+    size_t domain_id = node->get_node_options().context()->get_domain_id();
+    auto domain_id_topic_pair = std::make_pair(domain_id, topic_name);
+    if (map.find(domain_id_topic_pair) != map.end()) {
+      throw BridgeExistsError(
+              node->get_node_options().context()->get_domain_id(),
+              topic_name
+      );
```
ivanpauno, line 107: I'm not sure if throwing an error here is the best thing to do

ivanpauno (approved):
Seems like a good start.

I've left some comments, but you can ignore them here and address them in the next iteration if you want.

package.xml
```diff
@@ -0,0 +1,22 @@
...
+  <version>0.0.0</version>
+  <description>ROS 2 Domain Bridge</description>
+  <maintainer email="jacob@openrobotics.org">Jacob Perron</maintainer>
+  <license>Apache 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <depend>rclcpp</depend>
+  <depend>rcutils</depend>
+  <depend>rosbag2_cpp</depend>
```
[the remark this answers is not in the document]
jacobperron: I'm hoping that we can drop it. I plan to test out ros2/rclcpp#1452 to see what changes might be necessary here (or there).

doc/design.md
```diff
@@ -0,0 +1,156 @@
...
+
+The bridge should work with any supported ROS type (e.g. .msg, .srv, .action, or .idl).
+Furthermore, the bridge should not have to declare any explicit dependencies on interfaces types (e.g. in code or package.xml).
+Rather, the bridge should dynamically load type support at runtime (and produce an error if support for the requested type does not exist).
+
+5. Accurately bridge Quality of Service settings
+
+All ROS communication entities (e.g. publishers and service clients) have [Quality of Service (QoS)](http://design.ros2.org/articles/qos.html) settings associated with them.
+The bridge should faithfully map the QoS settings of data coming from one domain into another domain.
+For example, a publisher with reliability policy set to "best effort" should continue to publish as "best effort" in the other domains when bridged.
```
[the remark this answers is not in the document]
jacobperron: Good point, I meant to address that in the design doc but forgot. In order to handle any sort of edge cases, I think it's best to create a one-to-one mapping for each publisher being bridged. In your example, we would create two sus/pub pairs in the bridge. The first set would use "best effort" and the second set would use "reliable". This way we can hopefully match the exact QoS for each publisher being bridged. This will hopefully avoid any issues overflowing history due to insufficient history depth.

Leading me to another question: how should be handle history and depth policies? AFAIK, we can't query that info from the publisher endpoint, so maybe the best we can do is expose it as a config option in the bridge.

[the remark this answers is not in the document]
jacobperron: > Leading me to another question: how should be handle history and depth policies? AFAIK, we can't query that info from the publisher endpoint, so maybe the best we can do is expose it as a config option in the bridge.

hmm.. nevermind, after some experimentation it seems we can query history and depth after all. I guess I was lead to believe it wasn't possible due to the [this API not returning them](https://github.com/ros2/rclpy/blob/3e030c09a3fbb1a5c4ac1f7d827c1deffff5d242/rclpy/rclpy/topic_endpoint_info.py#L160).

doc/design.md
```diff
@@ -0,0 +1,156 @@
...
+
+3. Bridge ROS actions from one domain to another.
+
+The same as (1), but for ROS actions.
+
+4. No explicit dependencies on ROS interface types
+
+The bridge should work with any supported ROS type (e.g. .msg, .srv, .action, or .idl).
+Furthermore, the bridge should not have to declare any explicit dependencies on interfaces types (e.g. in code or package.xml).
+Rather, the bridge should dynamically load type support at runtime (and produce an error if support for the requested type does not exist).
```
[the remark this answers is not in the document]
jacobperron: Yeah, that sounds right. I've updated this point in d7797d6

doc/design.md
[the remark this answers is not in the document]
jacobperron: I've added to the section in d7797d6

doc/design.md
```diff
@@ -0,0 +1,156 @@
...
+
+All ROS communication entities (e.g. publishers and service clients) have [Quality of Service (QoS)](http://design.ros2.org/articles/qos.html) settings associated with them.
+The bridge should faithfully map the QoS settings of data coming from one domain into another domain.
+For example, a publisher with reliability policy set to "best effort" should continue to publish as "best effort" in the other domains when bridged.
+
+6. Remapping
+
+[Remapping](https://design.ros2.org/articles/static_remapping.html) is a generally useful concept in ROS.
+The bridge should support remapping topic, service, and action names when mapping the name from one domain to another.
+It is possible to bridge different topic names with the same name from different domains, so we should keep this in mind when it comes to remapping names.
```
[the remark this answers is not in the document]
jacobperron: Yes. I've added an example in d7797d6

include/domain_bridge/domain_bridge.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Start the bridge.
+  /**
+   * Start forwarding ROS traffic between domains.
+   * Only topics registered with \ref bridge_topic() will have their messages forwarded.
+   *
+   * This method blocks indefinitely.
+   */
+  DOMAIN_BRIDGE_PUBLIC
+  void spin();
```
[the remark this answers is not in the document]
jacobperron: Makes sense. Are there any examples or docs on recommended patterns for interacting the the executor?

doc/design.md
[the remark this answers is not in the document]
ivanpauno: > hmm.. nevermind, after some experimentation it seems we can query history and depth after all.

how? AFAIR that wasn't possible

include/domain_bridge/domain_bridge.hpp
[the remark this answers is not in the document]
ivanpauno: > Makes sense. Are there any examples or docs on recommended patterns for interacting the the executor?

Not that I know of, but either having a getter of the node (or the node base interface) or having an add_to_executor method sound like the most general ways of achieving it.

doc/design.md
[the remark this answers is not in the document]
jacobperron: See https://github.com/ros2/rclpy/pull/686
I'm proposing adding history and depth to the string output of endpoint info, which shows up in `ros2 topic info --verbose`. It seems to work locally, but perhaps it's because of shared memory transport or something?

I tested by running a talker,

    ros2 run demo_nodes_cpp talker

and then using `ros2topic` (with the mentioned PR):

    ros2 topic info --verbose /chatter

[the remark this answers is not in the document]
jacobperron: I don't know *how* it works, just reporting what I've observed.

include/domain_bridge/domain_bridge.hpp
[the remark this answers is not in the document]
jacobperron: New nodes are created every time a new topic is bridged, so it makes more sense to have an add_to_executor method. I've add this and a new TopicBridgeOptions class for passing around options like the callback group (59cb440)

Regarding the callback group, I'm confused about it's relationship with nodes, subscriptions, and executors. It looks like there's several ways that callback groups can be created and passed around :confused: We can't pass it to the subscription object since it has type `rclcpp::SubscriptionBase`. The only option I see at the moment is to keep a list of callback groups and add them to the executor in add_to_executor, but I don't see what this achieves if they aren't associated with the subscriptions. I've left a TODO in the code for now: 

https://github.com/ros2/domain_bridge/blob/59cb44047d4b6cbe95994c3f0769db3afde6560d/src/domain_bridge/domain_bridge.cpp#L151

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <map>
+#include <memory>
+#include <sstream>
+#include <string>
+#include <utility>
+
+#include "domain_bridge/exceptions.hpp"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/executors/multi_threaded_executor.hpp"
```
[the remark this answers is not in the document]
jacobperron: The executor has been moved outside the class now, and I've switched to the single threaded executor in the executable.

src/domain_bridge/domain_bridge.cpp
jacobperron: This was vestigial, and removed in the latest changes anyways.

doc/design.md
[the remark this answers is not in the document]
ivanpauno: does it work with all rmw implementations?

AFAIR, that information is not always available (e.g. I'm almost sure it's not in fastrtps).

src/domain_bridge/domain_bridge.cpp
```diff
@@ -150,8 +145,12 @@ class DomainBridgeImpl
     const std::string & topic,
     const std::string & type,
     size_t from_domain_id,
-    size_t to_domain_id)
+    size_t to_domain_id,
+    const TopicBridgeOptions & options)
   {
+    // TODO(jacobperron): Do something with options.callback_group()
```
ivanpauno, line 151: this can be passed to `create_publisher/create_subscription` through `PublisherOptions/SubscriptionOptions`.

include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -0,0 +1,70 @@
...
+{
+public:
+  /// Constructor.
+  /**
+   *  Default values:
+   *
+   *    - callback_group = nullptr (node's default)
+   */
+  DOMAIN_BRIDGE_PUBLIC
+  TopicBridgeOptions();
```
ivanpauno, line 38:
```suggestion
  TopicBridgeOptions() = default;
```

and delete the implementation in the source file

include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Destructor.
+  DOMAIN_BRIDGE_PUBLIC
+  virtual
+  ~TopicBridgeOptions() = default;
+
+  /// Copy constructor.
+  DOMAIN_BRIDGE_PUBLIC
+  TopicBridgeOptions(const TopicBridgeOptions & other);
```
ivanpauno, line 47:
```suggestion
  TopicBridgeOptions(const TopicBridgeOptions & other) = default;
```

and delete the implementation in the source file

include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Assignment operator.
+  DOMAIN_BRIDGE_PUBLIC
+  TopicBridgeOptions &
+  operator=(const TopicBridgeOptions & other);
```
ivanpauno, line 52:
```suggestion
  operator=(const TopicBridgeOptions & other) = default;
```

and delete the implementation in the source file

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,210 @@
...
+  using PublisherMap = std::map<std::pair<size_t, std::string>, std::shared_ptr<GenericPublisher>>;
+  using NodeMap = std::map<size_t, std::shared_ptr<rclcpp::Node>>;
+
+  DomainBridgeImpl()
+  {
+  }
+
+  ~DomainBridgeImpl()
+  {
+  }
```
ivanpauno, line 52:
```suggestion
  DomainBridgeImpl() = default;

  ~DomainBridgeImpl() = default;
```

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,210 @@
...
+  /// Map of (domain_id, topic name) to subscriptions
+  SubscriptionMap subscription_map_;
+};  // class DomainBridgeImpl
+
+DomainBridge::DomainBridge()
+: impl_(std::make_unique<DomainBridgeImpl>())
+{}
+
+DomainBridge::~DomainBridge()
+{}
```
ivanpauno, line 193:
```suggestion
DomainBridge::~DomainBridge() = default;
```

iiuc you cannot move this to the header file because `DomainBridgeImpl` was forward declared there, but `= default` is still valid in the source file.

include/domain_bridge/domain_bridge.hpp
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * When the executor is running ROS traffic will be bridged between domains.
+   * Only topics registered with \ref bridge_topic() will have their messages forwarded.
+   *
+   * This method will add all nodes associated with this bridge to the provided executor.
+   *
+   * \param executor: The executor to add this domain bridge to.
+   */
+  DOMAIN_BRIDGE_PUBLIC
+  void add_to_executor(std::shared_ptr<rclcpp::Executor> executor);
```
ivanpauno, line 56:
```suggestion
  void add_to_executor(rclcpp::Executor & executor);
```

no need of a shared ptr here

doc/design.md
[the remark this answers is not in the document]
jacobperron: I checked again, it looks like we can't get that information from fastrtps or connext.

src/domain_bridge/domain_bridge.cpp
jacobperron: That's what I thought, but we aren't calling rclcpp's `create_publisher/create_subscription` and instead using the `rclcpp::PublisherBase` and `rclcpp::SubscriptionBase` directly, which only accepts the `rcl` options struct.

include/domain_bridge/exceptions.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace domain_bridge
+{
+
+class BridgeExistsError : public std::runtime_error
+{
+public:
+  DOMAIN_BRIDGE_PUBLIC
+  BridgeExistsError(size_t domain_id, const std::string topic)
+  : std::runtime_error(
+      "the topic '" + topic + "' is already bridged from domain " + std::to_string(domain_id))
```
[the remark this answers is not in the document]
jacobperron: I think it *should* be possible, but it isn't at the moment. I'll work on it.

src/domain_bridge/domain_bridge.cpp
jacobperron: I suppose we could print a warning instead.

src/domain_bridge/domain_bridge.cpp
jacobperron: I've made this ID configurable via a new options structure: b39167a

This way users can get a deterministic output for the node names.

include/domain_bridge/domain_bridge.hpp
jacobperron: a559080

src/domain_bridge/domain_bridge.cpp
jacobperron: a559080

src/domain_bridge/domain_bridge.cpp
jacobperron: a559080

include/domain_bridge/topic_bridge_options.hpp
jacobperron: a559080

include/domain_bridge/topic_bridge_options.hpp
jacobperron: a559080

include/domain_bridge/topic_bridge_options.hpp
jacobperron: a559080

jacobperron:
I want to write some unit tests for this, but I'll add them in a separate PR for easier review.

src/domain_bridge/domain_bridge.cpp
jacobperron: Nevermind, I see where I can pass the callback group. It must have been too late in the day yesterday :expressionless:

jacobperron: Done a290340

src/domain_bridge/domain_bridge.cpp
jacobperron: Printing a warning instead: fbabc03

include/domain_bridge/exceptions.hpp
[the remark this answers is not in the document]
jacobperron: I've refactored so we can handle this case now: fbabc03

doc/design.md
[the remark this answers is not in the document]
jacobperron: I've added a section to the design doc discussing this point: c420562

jacobperron:
@ivanpauno I think I've addressed all your comments. Please take another look :bow: 

@mjeronimo It would be great if you could take a look over the design document before I continue as well :slightly_smiling_face:

src/domain_bridge/generic_publisher.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+namespace domain_bridge
+{
+
+GenericPublisher::GenericPublisher(
+  rclcpp::node_interfaces::NodeBaseInterface * node_base,
+  const rosidl_message_type_support_t & type_support,
+  const std::string & topic_name,
+  const rclcpp::QoS & qos)
+: rclcpp::PublisherBase(node_base, topic_name, type_support, rosbag2_get_publisher_options(qos))
```
ivanpauno, line 42: the problem with callback groups is that this generic publisher class doesn't take a `PublisherOptions` argument and instead uses `rosbag2_get_publisher_options(qos)`.

I guess https://github.com/ros2/rclcpp/pull/1452 will fix that

include/domain_bridge/domain_bridge_options.hpp
```diff
@@ -0,0 +1,71 @@
...
+  /// Set the identifier for the domain bridge.
+  /**
+   * The identifier is added to names of nodes created by the domain bridge.
+   */
+  DOMAIN_BRIDGE_PUBLIC
+  DomainBridgeOptions &
+  id(std::size_t id);
+
+private:
+  std::size_t id_{0u};
```
ivanpauno, line 66: This should IMO be generated by the implementation and not specifed by the user.

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,218 @@
...
+    const std::string & topic_name,
+    const rosidl_message_type_support_t & typesupport_handle,
+    rclcpp::CallbackGroup::SharedPtr group)
+  {
+    auto publisher = std::make_shared<GenericPublisher>(
+      node->get_node_base_interface().get(),
+      typesupport_handle,
+      topic_name,
+      rclcpp::QoS(10));
+    node->get_node_topics_interface()->add_publisher(publisher, std::move(group));
```
ivanpauno, line 100: this doesn't look fine IMO.
The `GenericPublisher` implementation should already be calling this internally

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,218 @@
...
+      typesupport_handle,
+      topic_name,
+      rclcpp::QoS(10),
+      [publisher](std::shared_ptr<rclcpp::SerializedMessage> msg) {
+        // Publish message into the other domain
+        auto serialized_data_ptr = std::make_shared<rcl_serialized_message_t>(
+          msg->get_rcl_serialized_message());
+        publisher->publish(serialized_data_ptr);
+      });
+    node->get_node_topics_interface()->add_subscription(subscription, std::move(group));
```
ivanpauno, line 123: same

src/domain_bridge/domain_bridge.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "topic_bridge.hpp"
+
+namespace domain_bridge
+{
+
+/// Implementation of \ref DomainBridge.
+class DomainBridgeImpl
+{
+public:
+  using NodeMap = std::map<std::size_t, std::shared_ptr<rclcpp::Node>>;
```
ivanpauno, line 45:
```suggestion
  using NodeMap = std::unordered_map<std::size_t, std::shared_ptr<rclcpp::Node>>;
```

ivanpauno (approved):
Some extra comments, but still lgtm

include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Assignment operator.
+  DOMAIN_BRIDGE_PUBLIC
+  TopicBridgeOptions &
+  operator=(const TopicBridgeOptions & other) = default;
+
+  /// Get callback group associated with the topics pub/sub pair.
+  DOMAIN_BRIDGE_PUBLIC
+  std::shared_ptr<rclcpp::CallbackGroup>
+  callback_group() const;
```
mjeronimo, line 57: Having read this far, I'm not clear on the usage of this callback_group. I assume its part of the implementation and not for the user.

src/domain_bridge.cpp
```diff
@@ -0,0 +1,36 @@
...
+{
+  rclcpp::init(argc, argv);
+
+  // TODO(jacobperron): Configure via YAML file
+  domain_bridge::DomainBridge domain_bridge;
+  domain_bridge.bridge_topic("image", "sensor_msgs/msg/Image", 5, 10);
+
+  rclcpp::executors::SingleThreadedExecutor executor;
+  domain_bridge.add_to_executor(executor);
+  executor.spin();
```
mjeronimo, line 31: Can topics/services/actions be added to the bridge dynamically, based on changing conditions? Or, is ti required for the bridge to be completely configured first via YAML or bridge_topic calls and then run?

src/domain_bridge/generic_publisher.hpp
```diff
@@ -0,0 +1,46 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
```
mjeronimo, line 1: Will these be factored out and made available to both the domain bridge and rosbag2? Seems like generally useful functionality.

src/domain_bridge/domain_bridge.cpp
jacobperron: That makes sense, I could move this inside the GenericPublisher. But it shouldn't break anything doing it here right? It just makes more sense to encapsulate it in the GenericPublisher.

jacobperron: I guess this will be fixed when switching to the upstream GenericPublisher too.

ivanpauno: > But it shouldn't break anything doing it here right?

I thought it was going to break something because I imagined that `rclcpp::PublisherBase` was already doing that, but it seems that's not the case.
but `rclcpp::create_publisher` is currently doing this.

> It just makes more sense to encapsulate it in the GenericPublisher.

:+1: yes

jacobperron: I just copied this implementation from rosbag2_transport, but hopefully the rclcpp version will do this encapsulation.

doc/design.md
```diff
@@ -0,0 +1,168 @@
...
+
+A C++ library with a public API is provided for users to call in their own process and extend as they like.
+The public API is expected to evolve over time, but at the very least users should be able to bridge ROS networks primitives
+
+For convenience, a standalone binary executable is also provided for easy integration into ROS systems.
+
+### Supporting generic types
+
+When bridging data, serialized data will be read from one domain and forwarded to another domain.
+As long as the bridge does not need access to the typed data, there is no need for deserialization and, therefore, C++ type support.
```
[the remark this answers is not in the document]
jacobperron: Good question! I hadn't thought about how ROS 2 security might play into the domain bridge. AFAIK, ROS 2 security is configured by specifying rules for which node can communicate over what topics (and in what direction). I *think* this means that if someone wants to enable security with a domain bridge in their system, they would need to account for the nodes being created by the bridge itself, otherwise the bridge simply won't work. I'm also not sure how security deals with different domains, I expect one might have to define separate security policies for each domain ID. It's likely if security is *not* enabled for one half of the bridge, you are exposing the bridged data.

I'll do some investigation and add a section to the design doc about security.

include/domain_bridge/domain_bridge.hpp
```diff
@@ -0,0 +1,83 @@
...
+{
+
+class DomainBridgeImpl;
+
+/// Bridge ROS entities across different domains.
+class DomainBridge
+{
+public:
+  /// Constructor.
+  DOMAIN_BRIDGE_PUBLIC
```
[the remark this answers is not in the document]
jacobperron: Exactly, duplication vs package dependency.  Here is some past discussion about it:

https://discourse.ros.org/t/add-visibility-control-h-to-the-api-as-standard-usage/11681

and here is a PR that adds a common header to rcutils:

https://github.com/ros2/rcutils/pull/194

So, I think in practice we could use that header (still some boiler plate involved though, IIUC).

include/domain_bridge/domain_bridge_options.hpp
jacobperron: This was added to avoid duplicate node name in cases where the user creates two or more domain bridges. I think giving the user a way to identifier the bridge nodes is better so they can deterministically predict the names. If we randomly generate node names, then I imagine this would make it more difficult to enable ROS security.

I think this is kind of equivalent to giving a node a name. Maybe renaming the option to "name" makes more sense, what do you think?

jacobperron: And the generated node names could be "{options.name()}_{domain_id}".

ivanpauno: That convinces me, an user provided name sounds fine.

include/domain_bridge/domain_bridge.hpp
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * When the executor is running ROS traffic will be bridged between domains.
+   * Only topics registered with \ref bridge_topic() will have their messages forwarded.
+   *
+   * This method will add all nodes associated with this bridge to the provided executor.
+   *
+   * \param executor: The executor to add this domain bridge to.
+   */
+  DOMAIN_BRIDGE_PUBLIC
+  void add_to_executor(rclcpp::Executor & executor);
```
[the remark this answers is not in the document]
jacobperron: It's certainly not required to be this way.

The first draft contained it's own executor as an implementation detail, but I refactored after this comment: https://github.com/ros2/domain_bridge/pull/1#discussion_r583707497

I think this API makes the domain_bridge more flexible as a library component of a larger executable. Basically, an executor is needed by the bridge to process subscription callbacks, but choosing the type of executor seems arbitrary.
This API gives users more control over the execution model and how the library is integrated; it lets them tailor the bridge for their application. Compare this with the alternative of providing a single "spin" method, where users are overly constrained.

For users that don't care about choosing an executor, they can use the executable instead, where we've arbitrarily chosen a single threaded executor.

---

I can certainly better document the programming model in the API docs. I hadn't considered documenting the it in the design doc, which is a bit more high-level, but I'm happy to add a section there too if you think it would help.

src/domain_bridge.cpp
jacobperron: If the user instantiates their own instance of `DomainBridge`, I think they should be able to bridge things dynamically using the provided C++ API. This makes me think that we probably want "remove_bridge" API too.

For the executable, I was thinking of keeping it simple and only allowing users to configure it via a YAML at runtime.

src/domain_bridge/generic_publisher.hpp
jacobperron: Yes. See https://github.com/ros2/rclcpp/pull/1452

Once that lands I'll remove the duplicate logic in this repo.

include/domain_bridge/topic_bridge_options.hpp
jacobperron: This is meant as a configuration point for the user. This relates to the executor discussion above https://github.com/ros2/domain_bridge/pull/1#discussion_r586759175. It enables a more advance users to customize the execution model.

I work on improving the docs.

doc/design.md
```diff
@@ -0,0 +1,168 @@
...
+TODO
+
+### Configuration
+
+The [YAML](https://yaml.org/) language was chosen as an easy way to externally configure the domain bridge.
+YAML is relatively simple to read for humans and computers alike and is commonly used in other places of the ROS ecosystem (e.g. ROS parameters).
+
+The bridge will look for three keys, `topics`, `services`, and `actions`, each of which contains a list of objects.
+The key for each object is the input name of the ROS entity (topic, service, or action name).
+For each ROS entity, you must specifiy the `type` of the interface, the domain ID where input data will come from (`from_domain`), and the domain ID for where to bridge (`to_domain`).
```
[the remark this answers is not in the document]
jacobperron: In theory, I think we could drop the `type`. But in practice, in my limited understanding, we still need the type for loading some rmw support. I could dig deeper, but I'm inclined to wait and see how https://github.com/ros2/rclcpp/pull/1452 plays out, since that's where the `type` requirement is coming from.

src/domain_bridge/domain_bridge.cpp
jacobperron: Updated 43e484d

include/domain_bridge/domain_bridge_options.hpp
jacobperron: Refactored to use a string, `name`, instead of an ID (2379dab)

doc/design.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Requirements
+
+The following are a set of requirements we would like our domain bridge to meet:
+
+1. Bridge ROS topics from one domain to another.
+
+This is pretty self-explanatory.
+Given a node *A* on domain *D<sub>A</sub>* and a node *B* on domain *D<sub>B</sub>*, where D<sub>A</sub> != D<sub>B</sub>,
+A and B should be able to communicate with the ROS pub/sub mechanism over one or more topics.
```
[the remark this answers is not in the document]
jacobperron: I've tried to clarify things in 5b8584c

doc/design.md
```diff
@@ -0,0 +1,168 @@
+# domain_bridge Design
+
+This article documents the design of the domain_bridge.
+
+## Background
+
+A ROS 2 network can be partitioned by assigning each *node* (more specifically, a *context* associated with the node), a *domain ID*.
```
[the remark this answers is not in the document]
jacobperron: Addressed in 5b8584c

doc/design.md
```diff
@@ -0,0 +1,168 @@
...
+
+For example, users may want to remap topic `/foo` in domain ID `1` to a different name `/bar` in domain ID `2`.
+
+7. Easily configurable
+
+Since the set of bridged data depends on the application, we need an easy way to configure what data is bridged.
+Users should be able to specify the following information at an API level:
+
+- The domain IDs that should be bridged
+- The topic names and types that should be bridged (same for services and actions)
```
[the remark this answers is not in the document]
jacobperron: Thanks for the suggestion. I've tried to make things more topic-oriented in 5b8584c

doc/design.md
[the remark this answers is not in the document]
jacobperron: I've started a section on security in the design doc (253aeb9). I *think* security will "just work", but we may need to document some best practices for enabling it. I'll come back and fill in the TODO in a follow-up PR once we've worked out a solution and document it.

include/domain_bridge/domain_bridge_options.hpp
```diff
@@ -0,0 +1,71 @@
...
+  DomainBridgeOptions(const DomainBridgeOptions & other) = default;
+
+  /// Assignment operator.
+  DOMAIN_BRIDGE_PUBLIC
+  DomainBridgeOptions &
+  operator=(const DomainBridgeOptions & other) = default;
+
+  /// Get the identifier for the domain bridge.
+  DOMAIN_BRIDGE_PUBLIC
+  std::size_t
```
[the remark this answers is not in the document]
jacobperron: Refactored to `std::string` in 2379dab

include/domain_bridge/domain_bridge.hpp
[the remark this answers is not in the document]
jacobperron: I've updated the API docs a little bit: 48bf2f1

doc/design.md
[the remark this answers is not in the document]
mjeronimo: Yes, that helps, thanks.

doc/design.md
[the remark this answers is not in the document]
mjeronimo: Great. Good to document the assumptions and implications as we understand them.

doc/design.md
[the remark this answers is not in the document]
mjeronimo: Sure.

include/domain_bridge/topic_bridge_options.hpp
mjeronimo: That would be helpful. I'd like to understand how one could use this feature. For example, could one implement some kind of filtering proxy model using this? Could log messages that have been bridged, potentially block messages (although I'm not sure why one would do this). Looking for some ideas on how it would be used.

src/domain_bridge.cpp
mjeronimo: Yes, that sounds reasonable to me.

src/domain_bridge/generic_publisher.hpp
mjeronimo: Great.

jacobperron:
@ivanpauno @mjeronimo Thanks for the reviews :bow: 
I think I've address most comments. I've got a local list of follow-ups that I'll chip away at based on feedback so far. I'm going to merge this, but I'm happy to address any more comments in follow-up PRs.

PR #2, opened 2021-03-05
Add unit tests and CI

jacobperron:
Adding unit tests and CI (for posterity).

### DIFF
--- a/test/domain_bridge/test_domain_bridge.cpp
+++ b/test/domain_bridge/test_domain_bridge.cpp
@@ -1,0 +1,179 @@
+// Copyright 2021, Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <string>
+
+#include "rclcpp/context.hpp"
+#include "rclcpp/executors/single_threaded_executor.hpp"
+
+#include "domain_bridge/domain_bridge.hpp"
+
+class TestDomainBridge : public ::testing::Test
+{
+protected:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+
+  static void TearDownTestCase()
+  {
+    rclcpp::shutdown();
+  }
+};
+
+TEST_F(TestDomainBridge, construction_destruction)
+{
+  // Default
+  {
+    domain_bridge::DomainBridge bridge;
+  }
+  // With options
+  {
+    domain_bridge::DomainBridge bridge(domain_bridge::DomainBridgeOptions());
+  }
+}
+
+TEST_F(TestDomainBridge, get_options)
+{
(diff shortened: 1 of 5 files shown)


test/domain_bridge/test_domain_bridge.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+TEST_F(TestDomainBridge, bridge_topic_invalid)
+{
+  // Invalid topic name
+  {
+    domain_bridge::DomainBridge bridge;
+    EXPECT_THROW(
+      bridge.bridge_topic("Not a v@lid topic name!", "test_msgs/msg/BasicTypes", 1, 2),
+      rclcpp::exceptions::InvalidTopicNameError);
```
jacobperron, line 109: I'd like to catch and throw exceptions defined in the `domain_bridge` namespace, for all invalid cases here that are throwing.
I'll do that in a separate PR if that sounds good to others.

mjeronimo: That's an interesting question, in general. The benefits of doing it that way include having a nice, clean abstract interface which includes domain_bridge-specific exceptions (only). However, with C++, if one relies on other libraries in the implementation, it's very hard to know which exceptions might be thrown. So, there's no guarantee to the user of this class that the provided exceptions are the only ones that might be thrown (unless you do something like catch (...)). Also, another drawback is having to catch exceptions in the implementation and then throwing semantically equivalent exceptions in the domain bridge namespace. I've typically approached catching exceptions with the idea that one should catch an exception if the code at that point can do something about the exception, otherwise, let it pass and maybe someone else along the call stack will handle it. So, I'm kind of ambivalent about which way to go. I lean towards your recommendation, but if we go that way, I think we should do this consistently with other ROS2 modules.

jacobperron: Good points. My initial thought was to provide a cleaner interface as you describe, but it's true it's difficult to account for all possible exceptions thrown by the underlying libraries. At the very least, I think we can make the interface a bit tidier for common/obvious error cases (like invalid topic or type names).

mjeronimo: Yes, agreed.

ivanpauno: IMHO, not catching these `rclcpp` exceptions is ok, they have pretty clear meaning already

test/domain_bridge/test_domain_bridge.cpp
```diff
@@ -0,0 +1,179 @@
...
+  {
+    domain_bridge::DomainBridge bridge;
+    EXPECT_THROW(bridge.bridge_topic("foo", "not a valid message type", 1, 2), std::runtime_error);
+  }
+  // Same bridge twice
+  {
+    testing::internal::CaptureStderr();
+    domain_bridge::DomainBridge bridge;
+    bridge.bridge_topic("foo", "test_msgs/msg/BasicTypes", 1, 2);
+    bridge.bridge_topic("foo", "test_msgs/msg/BasicTypes", 1, 2);
```
ivanpauno, line 121: shouldn't this throw an exception?

ivanpauno (approved):

test/domain_bridge/test_domain_bridge.cpp
jacobperron: Bridging the same topic/type for the same domains is essentially a noop. I originally was throwing an exception, though after your suggestion in https://github.com/ros2/domain_bridge/pull/1#discussion_r583752870, I switched to a warning, which seems reasonable to me since I don't think it is going to cause any severe errors.

ivanpauno: Sounds fine to me.
I think my original comment was because the first implementation didn't support bridging the same topic to two different domains, but I don't mind much.

jacobperron:
Cleaned up the commits to keep tests and CI as separate changes.

PR #3, opened 2021-03-08
Fix topic bridge less operator

jacobperron:
The first commit fixes a bug: f6dedb6
The second commit refactors the less-than logic as an operator of the struct: 40a4ffd

### DIFF
--- a/src/domain_bridge/topic_bridge.hpp
+++ b/src/domain_bridge/topic_bridge.hpp
@@ -58,20 +58,20 @@ struct TopicBridge
     if (from_domain_id < other.from_domain_id) {
       return true;
     }
-    if (from_domain_id != other.from_domain_id) {
+    if (from_domain_id > other.from_domain_id) {
       return false;
     }
     if (to_domain_id < other.to_domain_id) {
       return true;
     }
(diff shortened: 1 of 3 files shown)


src/domain_bridge/topic_bridge.hpp
```diff
@@ -45,27 +45,37 @@ struct TopicBridge
...
     }
-    if (lhs.to_domain_id < rhs.to_domain_id) {
+    if (from_domain_id != other.from_domain_id) {
+      return false;
+    }
+    if (to_domain_id < other.to_domain_id) {
       return true;
     }
-    if (lhs.topic_name.compare(rhs.topic_name) < 0) {
+    if (to_domain_id != other.to_domain_id) {
```
mjeronimo, line 67: For clarity, I would've used ">" here; it maintains the consistency with the less than instead of switching gears to !=.

mjeronimo (approved):

src/domain_bridge/topic_bridge.hpp
jacobperron: Sounds good to me: 37f1288

PR #4, opened 2021-03-09
Support for configuring domain bridge with YAML

jacobperron:
This PR adds a new free function for creating and configuring the bridge from a YAML file.

In order to facilitate testing, a new API for getting a list of bridged topics was also added.

This PR also bumps the required version of C++ to C++17 to take advantage of the filesystem utility. Alternatively, we could fall back to `rcpputils` if C++17 turns out to be an issue.


src/domain_bridge/domain_bridge_from_yaml.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <string>
+
+#include "domain_bridge/domain_bridge.hpp"
+#include "domain_bridge/domain_bridge_from_yaml.hpp"
+#include "domain_bridge/domain_bridge_options.hpp"
+
+namespace domain_bridge
+{
+
+std::unique_ptr<DomainBridge> domain_bridge_from_yaml(std::string file_path)
```
ivanpauno, line 28: Some suggestions about the API, feel free to ignore them:

- Why returning an `std::unique_ptr<>`? Returning a `DomainBridge` directly sounds better to me.
  You might need to define a move constructor/assignment operator (`= default` will be fine).
- I would rather have a `std::pair<DomainBridgeOptions, std::vector<std::pair<TopicBridge, TopicBridgeOptions>>> parse_domain_bridge_yaml_config(std::filesystem::path file_path)` function, and then a `domain_bridge_from_yaml()` that reuses it. And I would rather define a `struct DomainBridgeConfig` instead of the `std::pair<...>` thing I proposed.
- I would rather use `std::filesystem::path file_path` than `std::string` (or the rcpputils one if we prefer it).

test/domain_bridge/test_domain_bridge.cpp
```diff
@@ -60,42 +62,53 @@ TEST_F(TestDomainBridge, get_options)
...
   // This use-case isn't clearly supported by ROS 2, see https://github.com/ros2/ros2/issues/1095
   // {
   //   domain_bridge::DomainBridge bridge;
-  //   bridge.bridge_topic("foo", "test_msgs/msg/BasicTypes", 1, 2);
-  //   bridge.bridge_topic("foo", "test_msgs/msg/Nested", 1, 2);
-  //   bridge.bridge_topic("foo", "test_msgs/msg/Constants", 1, 2);
+  //   bridge.bridge_topic("foo", "test_msgs/msg/BasicTypes", 1u, 2u);
+  //   bridge.bridge_topic("foo", "test_msgs/msg/Nested", 1u, 2u);
+  //   bridge.bridge_topic("foo", "test_msgs/msg/Constants", 1u, 2u);
   // }
```
ivanpauno, line 112: this test case can just be removed

ivanpauno (approved):

test/domain_bridge/test_domain_bridge.cpp
jacobperron: I want to document that this undefined behavior (or maybe simply guard against it by declaring it invalid). I'll address this is a follow up PR

src/domain_bridge/domain_bridge_from_yaml.cpp
jacobperron: Thanks for the suggestions!
I've added a move constructor/assignment operator to avoid a unique_ptr in 46e10c7.
I've also changed the parameter type to `std::filesystem::path` 3b9bcc0.

I'll defer breaking the implementation into two functions, unless there's a strong reason to do it. It seems like more overhead than it's worth at the moment; and could easily be done in the future without breaking things.

ivanpauno (approved):

src/domain_bridge/domain_bridge.cpp
```diff
@@ -125,24 +129,16 @@ class DomainBridgeImpl
...
+    const std::string & topic = topic_bridge.topic_name;
+    const std::string & type = topic_bridge.type_name;
+    const std::size_t & from_domain_id = topic_bridge.from_domain_id;
+    const std::size_t & to_domain_id = topic_bridge.to_domain_id;
 
     // Check if already bridged
-    auto find_result = bridged_topics_.find(topic_bridge);
-    if (find_result != bridged_topics_.end()) {
+    if (bridged_topics_.find(topic_bridge) != bridged_topics_.end()) {
       std::cerr << "Topic '" << topic << "' with type '" << type << "'" <<
```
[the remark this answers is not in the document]
jacobperron: I'll look into using a ROS logger here. Alternatively, we could throw an exception, which is easy for an application to catch.

PR #5, opened 2021-03-15
Automatically match QoS settings across the bridge

jacobperron:
I've updated the design doc to reflect the implementation.

Specifically, this change introduces a `WaitForQosHandler` class used for deferring topic bridge creation. It creates a thread for each topic bridge to wait for at least one publisher to be available. Once a publisher is available it signals to the domain bridge via a callback, and a topic bridge is created.

In the special case of more than one publisher with different QoS settings, I've adopted a similar approach as rosbag2 for selecting a QoS that is compatible with most of the available publishers. Note, we could factor out this logic following https://github.com/ros2/rosbag2/issues/601 and/or https://github.com/ros2/rmw/issues/304.

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -43,9 +43,11 @@ All ROS communication entities (e.g. publishers and service clients) have [Quali
 The bridge should faithfully map the QoS settings of data coming from one domain into another domain.
 For example, a publisher with reliability policy set to "best effort" should continue to publish as "best effort" in the other domains when bridged.
 
-If there are multiple publishers on the same topic, but with different QoS settings, the bridge should have two streams of data, each with their own QoS settings.
+If there are multiple publishers on the same topic, but with different QoS settings, it would be nice if the bridge could preserve each stream of data.
 For example, if there is one publisher using "best effort" and another publisher using "reliable" on the same topic (with the same domain ID) and a bridge is made,
 then the bridge should forward data as "best effort" for the first publisher and as "reliable" for the second publisher into the output domain.
+Unforunately, this may not be possible due to technical limitations so we leave this as a soft requirement.
+Refer to the *QoS mapping* section of the proposed approach below for a more detailed discussion.
 
 Since remotely querying the *history* and *depth* QoS policies is not possible in many implementations (e.g. it is not required by the DDS spec),
 the bridge should offer a way to configure these values explicitly.
@@ -99,7 +101,6 @@ For example, the following diagram illustrates a configuration where the "/chatt
 A C++ library with a public API is provided for users to call in their own process and extend as they like.
 The public API is expected to evolve over time, but at the very least users should be able to bridge ROS networks primitives.
 
-
 For convenience, a standalone binary executable is also provided for easy integration into ROS systems.
 
 ### Supporting generic types
@@ -112,7 +113,41 @@ In fact, the generic publisher and subscription implementation is being moved to
 
 ### QoS mapping
 
-TODO
+The QoS settings of a publisher can be queried by the bridge.
+With this information, the bridge can create a subscription using the same QoS settings for receiving data.
+The bridge can also create a publisher for the domain being bridged with the same QoS settings.
+In this way, for the single publisher case, the QoS settings are preserved from one domain to another domain.
+
+We must consider the scenario when the domain bridge starts *after* the publisher of a bridged topic becomes available.
+In this case, the bridge cannot know what QoS settings to use for the bridged topic.
+The solution is to have the bridge wait until a publisher becomes available before creating it's own subscription and publisher for that topic.
+
+#### Multiple publishers
+
+As mentioned in the requirements, if there are multiple publishers on the same topic, it would be nice if the bridge could preserve multiple streams of data, each with their own QoS settings.
+Unforunately, this is technically challenging because it is difficult to associate a ROS message received by a subscription with the publisher that originally published the message.
+Consider the following scenario:
+
+1. Publisher *A* publishes on topic "chatter" with a QoS reliability setting of *reliable*.
+2. Publisher *B* publishes on topic "chatter" with a QoS reliability setting of *best effort*.
+3. In order to receive a message from *B*, we must create a subscription with a QoS reliability setting of *best effort*.
+4. The subscription will also receive messages from *A*, since *best effort* subscriptions also match with *reliable* publishers.
+
+If we cannot distinguish whether a message came from publisher *A* or publisher *B*, then we cannot know what QoS settings to use for the message when publishing into another domain.
+
+Instead, the proposed approach will do as best it can to ensure all messages make it across the bridge.
+The bridge will evaluate the QoS settings of all publishers and modifiy the QoS settings of the bridges subscription and publisher
+such that it matches the majority of the available publishers for a given topic.
+
+The bridge will decide on the QoS settings as soon as one or more publishers is available.
+This means it is possible that publishers joining after the topic bridge is created may have compatibility issues, and fail to have their messages bridged.
+For example, in the following scenario, messages coming from publisher *B* will not be bridged:
+
+1. Publisher *A* for topic "chatter" with a QoS reliability setting of *reliable* is created.
+2. A topic bridge is created for topic "chatter".
+3. Publisher *B* for topic "chatter" with a QoS reliability setting of *best effort* is created.
+
+We consider having multiple publishers with different QoS policies on the same topic to be rare in practice, and so do not try to handle the above scenario in the proposed approach.
 
 ### Remapping
 
@@ -181,3 +216,16 @@ Pros:
 Cons:
 - Requires generating/building SOSS type support for every interface being bridged.
 - Slightly less performant due to extra data marshaling.
+
+### Handling multiple publishers
+
+It's possible that the bridge could dynamically adjust to new publishers joining the network for an existing topic bridge.
+For example, the bridge could continuously monitor for new publishers and re-create it's subscription and publisher to reflect any necessary changes to QoS settings.
+Though, it is possible that some messages may be missed during the re-creation of the subscription.
+
+Pros:
+- Better suited to systems where publishers are created/destroyed during runtime
+
+Cons:
+- Not obvious how to handle missing out on messages during subscription re-creation
+- Substantially more complex to implement.
--- a/doc/design.md
+++ b/doc/design.md
@@ -117,10 +117,23 @@ The QoS settings of a publisher can be queried by the bridge.
 With this information, the bridge can create a subscription using the same QoS settings for receiving data.
 The bridge can also create a publisher for the domain being bridged with the same QoS settings.
 In this way, for the single publisher case, the QoS settings are preserved from one domain to another domain.
+However, there are issues related to the *liveliness* policy as well as multiple publishers per topic that are discussed below.
 
 We must consider the scenario when the domain bridge starts *after* the publisher of a bridged topic becomes available.
 In this case, the bridge cannot know what QoS settings to use for the bridged topic.
 The solution is to have the bridge wait until a publisher becomes available before creating it's own subscription and publisher for that topic.
+
+#### Liveliness
+
+The liveliness QoS policy can either be set to "automatic" or "manual by topic".
+If it is "automatic", the system will use the published messages as a heartbeat to consider if the publishers node is alive.
+If it is "manual by topic", the node needs to manually call a publisher API to assert it is alive.
+
+This poses a problem for the domain bridge.
+If the liveliness of a publisher is "manual by topic", then the bridge cannot mimic the QoS behavior into another domain.
+It would require the bridge to know when the original publisher is asserting its liveliness (i.e. calling the publisher API).
+Since there is no mechanism in ROS 2 to get this information at the subscription site, the bridge cannot support "manual by topic" liveliness.
+Therefore, the bridge will always use "automatic" liveliness, regardless of the original publishers policy.
 
 #### Multiple publishers
 
@@ -139,6 +152,9 @@ Instead, the proposed approach will do as best it can to ensure all messages mak
 The bridge will evaluate the QoS settings of all publishers and modifiy the QoS settings of the bridges subscription and publisher
 such that it matches the majority of the available publishers for a given topic.
 For example, given publisher *A* and publisher *B* from the aforementioned scenario, the bridge would select a reliability setting of best effort since it matches with both publishers.
+
+For *deadline* and *lifespan* policies, the bridge will use the max value given the policy value for all publishers.
+For example, if a publisher *A* has a larger deadline than publisher *B* (and *A* and *B* publish to the same topic), then the bridge will use the deadline value of publisher *A*.
 
 The bridge will decide on the QoS settings as soon as one or more publishers is available.
 This means it is possible that publishers joining after the topic bridge is created may have compatibility issues, and fail to have their messages bridged.
--- a/doc/design.md
+++ b/doc/design.md
@@ -131,13 +131,14 @@ Consider the following scenario:
 1. Publisher *A* publishes on topic "chatter" with a QoS reliability setting of *reliable*.
 2. Publisher *B* publishes on topic "chatter" with a QoS reliability setting of *best effort*.
 3. In order to receive a message from *B*, we must create a subscription with a QoS reliability setting of *best effort*.
-4. The subscription will also receive messages from *A*, since *best effort* subscriptions also match with *reliable* publishers.
+4. The subscription will also receive messages from *A*, since *best effort* subscriptions also matches with *reliable* publishers.
 
 If we cannot distinguish whether a message came from publisher *A* or publisher *B*, then we cannot know what QoS settings to use for the message when publishing into another domain.
 
 Instead, the proposed approach will do as best it can to ensure all messages make it across the bridge.
 The bridge will evaluate the QoS settings of all publishers and modifiy the QoS settings of the bridges subscription and publisher
 such that it matches the majority of the available publishers for a given topic.
+For example, given publisher *A* and publisher *B* from the aforementioned scenario, the bridge would select a reliability setting of best effort since it matches with both publishers.
 
 The bridge will decide on the QoS settings as soon as one or more publishers is available.
 This means it is possible that publishers joining after the topic bridge is created may have compatibility issues, and fail to have their messages bridged.
@@ -224,8 +225,8 @@ For example, the bridge could continuously monitor for new publishers and re-cre
 Though, it is possible that some messages may be missed during the re-creation of the subscription.
 
 Pros:
-- Better suited to systems where publishers are created/destroyed during runtime
+- Better suited to systems where publishers with different QoS settings are created/destroyed during runtime
 
 Cons:
-- Not obvious how to handle missing out on messages during subscription re-creation
+- Not obvious how to handle missing messages during subscription re-creation
 - Substantially more complex to implement.
--- a/doc/design.md
+++ b/doc/design.md
@@ -46,7 +46,7 @@ For example, a publisher with reliability policy set to "best effort" should con
 If there are multiple publishers on the same topic, but with different QoS settings, it would be nice if the bridge could preserve each stream of data.
 For example, if there is one publisher using "best effort" and another publisher using "reliable" on the same topic (with the same domain ID) and a bridge is made,
 then the bridge should forward data as "best effort" for the first publisher and as "reliable" for the second publisher into the output domain.
-Unforunately, this may not be possible due to technical limitations so we leave this as a soft requirement.
+Unfortunately, this may not be possible due to technical limitations so we leave this as a soft requirement.
 Refer to the *QoS mapping* section of the proposed approach below for a more detailed discussion.
 
 Since remotely querying the *history* and *depth* QoS policies is not possible in many implementations (e.g. it is not required by the DDS spec),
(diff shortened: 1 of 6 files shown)


src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+/// QoS match information.
+/**
+ * Used as a return value for QoS matching functions.
+ */
+struct QosMatchInfo
+{
+  explicit QosMatchInfo(const rclcpp::QoS & init_qos)
+  : qos(init_qos)
+  {}
```
ivanpauno, line 42:
```suggestion
  explicit QosMatchInfo()
  : qos{10}
  {}
```

see next comment.
(depth doesn't matter, I guess it's overriden by the user provided depth (if there is one).

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,206 @@
...
+      node->get_publishers_info_by_topic(topic);
+    std::size_t num_endpoints = endpoint_info_vec.size();
+
+    // If there are no publishers, return default QoS
+    if (num_endpoints < 1u) {
+      return {};
+    }
+
+    // Initialize QoS arbitrarily
+    QosMatchInfo result_qos(endpoint_info_vec[0].qos_profile());
```
ivanpauno, line 156:
```suggestion
    QosMatchInfo result_qos;
    result_qos.qos.reliability(endpoint_info_vec[0].qos_profile().reliability());
    result_qos.qos.durability(endpoint_info_vec[0].qos_profile().durability());
```

I think copying the liveliness and deadline of the read profile isn't a good idea.
Each of the publishers might have a different one, but the good thing is that if your subscription has the default liveliness and default deadline it will match everything.

ivanpauno (commented):
Looks like a good start, I've left some comments.

Note to self: review the test cases.

doc/design.md
```diff
@@ -112,7 +113,42 @@ In fact, the generic publisher and subscription implementation is being moved to
...
 
-TODO
+The QoS settings of a publisher can be queried by the bridge.
+With this information, the bridge can create a subscription using the same QoS settings for receiving data.
+The bridge can also create a publisher for the domain being bridged with the same QoS settings.
+In this way, for the single publisher case, the QoS settings are preserved from one domain to another domain.
+
+We must consider the scenario when the domain bridge starts *after* the publisher of a bridged topic becomes available.
+In this case, the bridge cannot know what QoS settings to use for the bridged topic.
+The solution is to have the bridge wait until a publisher becomes available before creating it's own subscription and publisher for that topic.
```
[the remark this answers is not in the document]
jacobperron: I'll add a QoS as an option to TopicBridgeOptions and expose as a YAML config in a follow-up https://github.com/ros2/domain_bridge/issues/8

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,206 @@
...
+          auto opt_qos = this->get_topic_qos(topic, node);
+          if (opt_qos) {
+            const QosMatchInfo & qos = opt_qos.value();
+            callback(qos);
+            return;
+          }
+        }
+      };
+    auto waiting_thread = std::make_shared<std::thread>(invoke_callback_when_qos_ready);
+    waiting_threads_.push_back(waiting_thread);
```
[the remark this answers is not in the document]
jacobperron: I considered this and meant to try a refactor to do it. I agree it would scale much better. I'll take a look at doing it.

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: Hmmm, yeah good point. We could also take the largest of all values for deadline and lifespan (note the QoS here is also applied to the publisher). Liveliness is trickier; I'm not sure how to handle the "manual by topic" since we are republishing to the other side. Maybe it's best to always use "automatic" for the bridge publisher. What do you think?

ivanpauno: >  I'm not sure how to handle the "manual by topic" since we are republishing to the other side

A subscription with "AUTOMATIC" liveliness will match everything, and the publisher should also be automatic if not you will need to have code doing some manual `assert_liveliness` calls.

> We could also take the largest of all values for deadline and lifespan

Sounds fine to me, that's in general the default but manuall setting the durations to "INFINITY" is better.

src/domain_bridge/domain_bridge.cpp
```diff
@@ -145,25 +170,47 @@ class DomainBridgeImpl
       return;
     }
 
+    // Create a null entry to avoid duplicate bridges
+    bridged_topics_[topic_bridge] = {nullptr, nullptr};
```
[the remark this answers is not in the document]
jacobperron: The condition at L166 above should be guarding against overriding existing entries. The reason for adding an entry here, and not in the callback, is to avoid multiple callbacks from adding the same entry.

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,206 @@
...
+  /// Register a callback that is called when QoS is ready for one or more publishers.
+  /**
+   * \param topic: The name of the topic to monitor.
+   * \param node: The node to use to monitor the topic.
+   * \param callback: User callback that is triggered when QoS settings can be determined
+   *   for the provided topic name.
+   *   If QoS settings are already available, then the callback is called immediately.
+   */
+  void register_on_publisher_qos_ready_callback(
+    std::string topic,
```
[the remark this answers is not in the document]
jacobperron: Making this `const &` seems like a premature optimization. I'm passing in a std::string in the one place this is used at the moment, and by leaving this as-is the compiler could take advantage of move semantics (though perhaps not with the current usage).

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,206 @@
...
+    // If the QoS is already avaiable, trigger the callback immediately
+    auto opt_qos = this->get_topic_qos(topic, node);
+    if (opt_qos) {
+      const QosMatchInfo & qos = opt_qos.value();
+      callback(qos);
+      return;
+    }
+
+    // Create a thread that waits for a publisher to become available
+    auto invoke_callback_when_qos_ready = [this, topic, node, callback]()
```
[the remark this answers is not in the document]
jacobperron: This introduces a bug, since the original `topic` variable goes out of scope before this function is run.

doc/design.md
```diff
@@ -43,9 +43,11 @@ All ROS communication entities (e.g. publishers and service clients) have [Quali
 The bridge should faithfully map the QoS settings of data coming from one domain into another domain.
 For example, a publisher with reliability policy set to "best effort" should continue to publish as "best effort" in the other domains when bridged.
 
-If there are multiple publishers on the same topic, but with different QoS settings, the bridge should have two streams of data, each with their own QoS settings.
+If there are multiple publishers on the same topic, but with different QoS settings, it would be nice if the bridge could preserve each stream of data.
 For example, if there is one publisher using "best effort" and another publisher using "reliable" on the same topic (with the same domain ID) and a bridge is made,
 then the bridge should forward data as "best effort" for the first publisher and as "reliable" for the second publisher into the output domain.
+Unforunately, this may not be possible due to technical limitations so we leave this as a soft requirement.
```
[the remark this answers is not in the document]
jacobperron: fd1342b

doc/design.md
```diff
@@ -112,7 +113,42 @@ In fact, the generic publisher and subscription implementation is being moved to
...
+In this way, for the single publisher case, the QoS settings are preserved from one domain to another domain.
+
+We must consider the scenario when the domain bridge starts *after* the publisher of a bridged topic becomes available.
+In this case, the bridge cannot know what QoS settings to use for the bridged topic.
+The solution is to have the bridge wait until a publisher becomes available before creating it's own subscription and publisher for that topic.
+
+#### Multiple publishers
+
+As mentioned in the requirements, if there are multiple publishers on the same topic, it would be nice if the bridge could preserve multiple streams of data, each with their own QoS settings.
+Unforunately, this is technically challenging because it is difficult to associate a ROS message received by a subscription with the publisher that originally published the message.
```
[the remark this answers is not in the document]
jacobperron: fd1342b

src/domain_bridge/domain_bridge.cpp
```diff
@@ -145,25 +170,47 @@ class DomainBridgeImpl
...
-
-    // Create subscription for the 'from_domain'
-    auto subscription = this->create_subscription(
-      from_domain_node, publisher, topic, *typesupport_handle, options.callback_group());
-
-    bridged_topics_[topic_bridge] = {publisher, subscription};
+    // Register a callback to be triggered when QoS settings are available for one or more
+    // publishers on the 'from' side of the bridge
+    // The callback may be triggered immediately if a publisher is available
+    auto create_bridge = [ = ](const QosMatchInfo & qos_match)
```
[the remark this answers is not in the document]
jacobperron: Yeah, I was just lazy :grin: 

Using explicit capture in 84ae94a

src/domain_bridge/domain_bridge.cpp
```diff
@@ -145,25 +170,47 @@ class DomainBridgeImpl
...
+      {
+        // Print any match warnings
+        for (const auto & warning : qos_match.warnings) {
+          std::cerr << warning << std::endl;
+        }
+
+        // Get typesupport handle
+        auto typesupport_handle = rosbag2_cpp::get_typesupport_handle(
+          type, "rosidl_typesupport_cpp", loaded_typesupports_[type]);
```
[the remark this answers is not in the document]
jacobperron: Good catch. Fixed in f1bb5c1

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+    // This implementation is inspired by rosbag2_transport:
+    // https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/qos.cpp
+
+    // Query QoS info for publishers
+    std::vector<rclcpp::TopicEndpointInfo> endpoint_info_vec =
+      node->get_publishers_info_by_topic(topic);
+    std::size_t num_endpoints = endpoint_info_vec.size();
+
+    // If there are no publishers, return default QoS
```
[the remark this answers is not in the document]
jacobperron: Updated c780649

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: > Sounds fine to me, that's in general the default but manuall setting the durations to "INFINITY" is better.

I'd like to try and mimic the QoS as best we can, so I'll avoid setting the durations to infinity by default.

I think we have to make an exception for liveliness, and explicitly use AUTOMATIC.

jacobperron: See c62b3a1

- Always use automatic liveliness
- Use max of all deadline
- Use max of all lifespan

Updated the design doc and added tests.

src/domain_bridge/wait_for_qos_handler.hpp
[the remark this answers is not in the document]
jacobperron: So, the problem is that the node used for querying topics is specific to the domain ID. In other words, I think we need at least one thread per domain ID. We could try to be a bit more clever and store a `std::map<node, std::vector<std::pair<topic, callback>>>` and have exactly one node (domain ID) per thread. The logic's going to get a bit more complex.

src/domain_bridge/wait_for_qos_handler.hpp
ivanpauno: > Use max of all deadline
> Use max of all lifespan

IMO, those two are better manually chosen, "replicating" them doesn't seem to completely make sense.

src/domain_bridge/wait_for_qos_handler.hpp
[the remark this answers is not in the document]
ivanpauno: That sounds fine to me, it's a bit tricky but considering that in the common case (bridging from one domain to one domain) you only have one thread it seems to be worth doing.

src/domain_bridge/wait_for_qos_handler.hpp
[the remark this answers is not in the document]
ivanpauno: sorry, my bad

src/domain_bridge/wait_for_qos_handler.hpp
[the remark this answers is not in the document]
ivanpauno: sounds fine to me, as you actually need to keep a copy of the string.
> and by leaving this as-is the compiler could take advantage of move semantics (though perhaps not with the current usage).

You have to actually explicitly call `std::move` to take advantage of it, but whatever.

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: They seem to make sense to me. If we have a publisher in domain A with a deadline, then using the same deadline for the bridge publisher on domain B will unsure subscriptions in domain B will experience similar behavior as subscriptions in domain A. Ditto for lifespan. What in particular doesn't make sense about this logic?

I will definitely expose a configuration point for users to override QoS values (e.g. via the YAML config). Expect a follow-up PR after this one.

ivanpauno: > They seem to make sense to me. If we have a publisher in domain A with a deadline, then using the same deadline for the bridge publisher on domain B will ensure subscriptions in domain B will experience similar behavior as subscriptions in domain A. 

The thing is that the bridge will introduce some delay, so you tipically won't be able to met the same deadline as the original publisher.
If deadlines are important, I would rather think what's a reasonable deadline for the new domain than letting the bridge infer them.

> Ditto for lifespan. What in particular doesn't make sense about this logic?

Lifespan is how much the message lives in the queue, and it's important how that's is combined with the history size/kind.
Because the last two cannot be instronspected, I don't think that inferring the lifespan will always lead to good results.

---

anyways, keeping this logic sounds fine to me.

jacobperron: > The thing is that the bridge will introduce some delay, so you tipically won't be able to met the same deadline as the original publisher.

Since deadline is a duration between messages, any delay introduced by the bridge should (hopefully) be systematic. Even though the total time from the original publisher to the endpoint in the another domain may take longer, the time between consecutive messages republished by the bridged should ideally remain the same as they were before they entered the bridge. E.g. if the original publisher is publishing at 10Hz, I would expect the bridge to also publish at 10Hz into the other domain.

> Lifespan is how much the message lives in the queue, and it's important how that's is combined with the history size/kind.

This is a good point. I can add a note about this potential pitfall in the design doc.

jacobperron:
So, the logic introduced in c62b3a1 uncovered an integer overflow bug in `rclcpp::Duration`. This PR depends on https://github.com/ros2/rclcpp/pull/1584 for QoS matching to work properly. Specifically, setting a deadline (or lifespan) with a value larger than INT32_MAX (which is the default value) doesn't work.

src/domain_bridge/wait_for_qos_handler.hpp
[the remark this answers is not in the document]
jacobperron: PTAL at 4ae1095

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,276 @@
...
+            if (topic_callback_vec.size() == 0u) {
+              waiting_map_.erase(waiting_map_.find(node));
+            }
+
+            // It's only worth continuing if we have callbacks to handle,
+            // or we're shutting down
+            cv->wait(
+              lock,
+              [this, &topic_callback_vec]
+              {return topic_callback_vec.size() > 0u || this->shutting_down_.load();});
```
ivanpauno, line 154: don't combine condition variables and atomics, this code is equivalent to:

```cpp
while (!(topic_callback_vec.size() > 0u || this->shutting_down_.load())) {
  // boolean flipped here and notification sent here, deadlock
  cv->wait(lock);
}
```

[this article](https://www.modernescpp.com/index.php/c-core-guidelines-be-aware-of-the-traps-of-condition-variables) explains the problem well.

tl;dr: replace the atomic with a normal `bool` and lock the `waiting_map_mutex_` before flipping it (no need to hold the mutex while calling `notify_all()` though).

ivanpauno: actually because of the logic used here, it seems that `topic_callback_vec.size() > 0u` is always true.
Maybe just deleting the condition variable code (?)

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -0,0 +1,276 @@
...
+        }
+      };
+    auto waiting_thread = std::make_shared<std::thread>(invoke_callback_when_qos_ready);
+    waiting_threads_[node] = {waiting_thread, cv};
+  }
+
+private:
+  using ThreadMap = std::unordered_map<
+    std::shared_ptr<rclcpp::Node>,
+    std::pair<std::shared_ptr<std::thread>, std::shared_ptr<std::condition_variable>>>;
```
ivanpauno, line 165: I don't mind, but it doesn't seem that you really need to wrap the `std::thread` and the `std::condition_variable` in a `std::shared_ptr` here

ivanpauno (commented):
looking good, only found some minor issues

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: I think without the condition variable, the thread essentially runs a busy-loop when `topic_callback_vec.size() == 0u`. There is no work to do, but it will continue looping and waiting for graph events.

jacobperron: Thanks for catching the deadlock bug! If I remove the atomic from the predicate, I get other shutdown issues; Looking into it.

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: `std::thread` and `std::condition_variable` are not copyable, so I think this is the only way I can store a list of them. Correct me if I'm wrong.

ivanpauno: > std::thread and std::condition_variable are not copyable, so I think this is the only way I can store a list of them. Correct me if I'm wrong.

you have to use `std::unordered_map::emplace()` to avoid that issue, e.g.:

```cpp
waiting_threads_.emplace(
  std::piecewise_construct,
  std::forward_as_tuple(node),
  std::forward_as_tuple(
    std::piecewise_construct, std::forward_as_tuple(invoke_callback_when_qos_ready), std::forward_as_tuple()));
```

so beautiful :joy:

src/domain_bridge/wait_for_qos_handler.hpp
ivanpauno: > I think without the condition variable, the thread essentially runs a busy-loop when topic_callback_vec.size() == 0u. There is no work to do, but it will continue looping and waiting for graph events.

True, I didn't check that correctly.

ivanpauno: > Thanks for catching the deadlock bug! If I remove the atomic from the predicate, I get other shutdown issues; Looking into it.

Just to be more clear, my recommendation was to replace this line

https://github.com/ros2/domain_bridge/blob/4ae1095a59b290fdd72ec68451ef66d76ab8afdc/src/domain_bridge/wait_for_qos_handler.hpp#L67

with

```cpp
{
  std::lock_guard<std::mutex> lock(waiting_map_mutex_);
  shutting_down_ = true; 
}
```

and `shutting_down_` can be made a `bool`.

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: Though it's perhaps slightly less performant, using shared_ptr seems easier to understand :sweat_smile:

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: Now using a regular bool for `shutting_down_` and fixed some cleanup logic in the event of a SIGINT: eac644d

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -120,38 +123,51 @@ class WaitForQosHandler
...
+                opt_qos = this->get_topic_qos(topic, node);
+              } catch (const rclcpp::exceptions::RCLError & ex) {
+                // If the context was shutdown, then exit cleanly
+                // This can happen if we get a SIGINT
+                const auto context = node->get_node_options().context();
+                if (!context->is_valid()) {
+                  return;
+                }
+                throw ex;
+              }
```
ivanpauno, line 156: argh... this doesn't look pretty nice

src/domain_bridge/wait_for_qos_handler.hpp
```diff
@@ -120,38 +123,51 @@ class WaitForQosHandler
...
-            }
 
-            // It's only worth continuing if we have callbacks to handle,
+            // It's only worth continuing if we have callbacks to handle
             // or we're shutting down
             cv->wait(
               lock,
-              [this, &topic_callback_vec]
-              {return topic_callback_vec.size() > 0u || this->shutting_down_.load();});
+              [this, &topic_callback_vec, event]
```
ivanpauno, line 169: why is this capturing `event`?

ivanpauno (approved):

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: Yeah, I know.. I think the only other way around it would be to disable the default signint handler and implement our own, but that wouldn't look pretty either.

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: It is vestigial. Removed: b3af55b

ivanpauno (approved):

src/domain_bridge/wait_for_qos_handler.hpp
jacobperron: I've reconsidered and changed the default values to be max integers instead of matching available publishers in https://github.com/ros2/domain_bridge/pull/13

PR #6, opened 2021-03-15
Refactor YAML parsing and allow default domain IDs

jacobperron:
- Change parsing function to return new struct DomainBridgeConfig.
- Add DomainBridge constructor that takes a DomainBridgeConfig.
- Support default domain IDs.


include/domain_bridge/parse_domain_bridge_yaml_config.hpp
```diff
@@ -33,42 +33,46 @@ class YamlParsingError : public std::runtime_error
...
 };
 
-/// Create a DomainBridge from a YAML file.
+/// Create a DomainBridgeConfig object from a YAML file.
 /**
  * The YAML file may contain the following optional keys:
  *
  * - name: Name of the domain bridge
+ * - from_domain: The default 'from_domain' used for bridged topics.
+ *     If omitted, then each topic bridge must specify it's own value.
```
mjeronimo, line 42: "it's" -> "its"

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
```diff
@@ -33,42 +33,46 @@ class YamlParsingError : public std::runtime_error
...
+ * - to_domain: The default 'to_domain' used for bridged topics
+ *     If omitted, then each topic bridge must specify it's own value.
  * - topics: A map of topic names to a map of topic bridge information
  *
- * For each topic bridge, the follow information is required, in the form of YAML keys:
+ * For each topic bridge, the follow keys are valid:
```
mjeronimo, line 47: "the follow" -> "the following"

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
```diff
@@ -33,42 +33,46 @@ class YamlParsingError : public std::runtime_error
...
  *
- * - type: Message type to bridge
- * - from_domain: Subscribe to the topic on this domain ID
- * - to_domain: Publish to the topic on this domain ID
+ * - type: Message type to bridge (required)
+ * - from_domain: Subscribe to the topic on this domain ID (optional)
```
mjeronimo, line 50: I don't know if I'd put "required" or "optional" here since it depends on the context - whether there is a default from_domain and/or to_domain.

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
```diff
@@ -33,42 +33,46 @@ class YamlParsingError : public std::runtime_error
...
  *
  * \param file_path: Path to the YAML file.
- * \return A pointer to an initialized DomainBridge that was configured based on input YAML file.
+ * \return A DomainBridgeConfig object populated according to the input YAML file.
  * \throws domain_bridge::YamlParsingError if there was an error parsing the YAML file,
  *   for example, if a required key is missing.
  */
 DOMAIN_BRIDGE_PUBLIC
-DomainBridge domain_bridge_from_yaml(std::filesystem::path file_path);
+DomainBridgeConfig parse_domain_bridge_yaml_config(std::filesystem::path file_path);
```
mjeronimo, line 74: Could also be a static method in a DomainBridgeYamlParser, like DomainBridgeYamlParser::parse().

jacobperron: I went with a free function since it isn't really stateful; what advantage does a `DomainBridgeYamlParser` object (or static method) have over this free function?

CMakeLists.txt
```diff
@@ -24,10 +24,10 @@ find_package(yaml_cpp_vendor REQUIRED)
 
 add_library(${PROJECT_NAME} SHARED
   src/${PROJECT_NAME}/domain_bridge.cpp
-  src/${PROJECT_NAME}/domain_bridge_from_yaml.cpp
   src/${PROJECT_NAME}/domain_bridge_options.cpp
   src/${PROJECT_NAME}/generic_publisher.cpp
   src/${PROJECT_NAME}/generic_subscription.cpp
+  src/${PROJECT_NAME}/parse_domain_bridge_yaml_config.cpp
```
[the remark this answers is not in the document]
jacobperron: Right. This is the name of the free function that is contained therein. I guess this relates to the question of whether this is a free-function or a class method: https://github.com/ros2/domain_bridge/pull/6#discussion_r595497012

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
mjeronimo: No advantage in terms of functionality. More along the lines of ensuring that we're all using a consistent approach. I have no issue with the particular name, per se, as long as similar kinds of functions are named in a similar way and don't use a different approach, like an object/static method. Personally, I would've thought of a parser object, as I mentioned, but there is no intrinsic benefit to that. Just making sure about the convention. 

Also, the consistency is for the reader of the code. It's nice, IMO, if there are predictable naming conventions.

jacobperron: I agree, we should be consistent. I'm still a little confused as to what we could change here to be consistent with other ROS 2 packages, though. Most free function names I see are actions, [for example](https://github.com/ros2/rclcpp/blob/35c89c8afcac89a43872f0bf5b763ff324c82b05/rclcpp/include/rclcpp/utilities.hpp). Perhaps renaming the header/cpp files as you suggest would help?

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
jacobperron: 609e2c9

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
jacobperron: 609e2c9

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
jacobperron: Removed qualifiers in 609e2c9

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
mjeronimo: Sorry, my fault for not being clear. I didn't mean that the code necessarily had to change, I just wanted to ensure that we were indeed using a consistent convention and that this was it. I raised the point because it's not what I would have done and wanted to check on this.

mjeronimo (approved):

include/domain_bridge/parse_domain_bridge_yaml_config.hpp
jacobperron: Gotcha. TBH, I picked the name sort-of arbitrarily. I'm not sure that we have an explicit convention written down anywhere. For style, we refer to the [Google style guide](https://google.github.io/styleguide/cppguide.html) (with some exceptions), and I didn't notice anything in there about this topic.

Issue #8, opened 2021-03-16
QoS overriding


jacobperron:
https://github.com/ros2/domain_bridge/pull/5 enables automatic QoS selection based on available publishers, but it would be nice to have the option to override the QoS used for a particular topic.

We could achieve this by adding to the `TopicBridgeOptions` class and exposing YAML configuration points.

In particular, it is already documented in the design doc that history and depth policies should be configurable since they can not be queried from publishers at runtime. Currently, they are defaulting to KEEP_LAST and 10.

PR #13, opened 2021-03-25
Add QoS overriding

jacobperron:
Resolves https://github.com/ros2/domain_bridge/issues/8

* Adds support for configuring QoS policies per topic bridge.
* Adds YAML support
* Change default value for deadline and lifespan policies to be "infinite"
  * Users can opt-in to the automatic matching

### DIFF
--- a/src/domain_bridge/domain_bridge.cpp
+++ b/src/domain_bridge/domain_bridge.cpp
@@ -16,6 +16,7 @@
 
 #include <cstddef>
 #include <iostream>
+#include <limits>
 #include <map>
 #include <memory>
 #include <sstream>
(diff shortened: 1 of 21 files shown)


src/domain_bridge/domain_bridge.cpp
```diff
@@ -185,6 +186,46 @@ class DomainBridgeImpl
...
+            qos.deadline(rclcpp::Duration::from_nanoseconds(deadline_ns));
+          }
+        } else {
+          qos.deadline(qos_match.qos.deadline());
+        }
+        if (qos_options.lifespan()) {
+          const auto lifespan_ns = qos_options.lifespan().value();
+          if (lifespan_ns < 0) {
+            qos.lifespan(
+              rclcpp::Duration::from_nanoseconds(std::numeric_limits<std::int64_t>::max()));
```
ivanpauno, line 218: maybe:

```suggestion
            qos.lifespan(RMW_DURATION_INFINITE);
```

we should have something like `rclcpp::Duration()::Infinity()` in rclcpp though.

I'm not sure if your code above does exactly the same, thought it should.

src/domain_bridge/domain_bridge.cpp
```diff
@@ -185,6 +186,46 @@ class DomainBridgeImpl
...
+        if (qos_options.durability()) {
+          qos.durability(qos_options.durability().value());
+        } else {
+          qos.durability(qos_match.qos.durability());
+        }
+        if (qos_options.deadline()) {
+          const auto deadline_ns = qos_options.deadline().value();
+          if (deadline_ns < 0) {
+            qos.deadline(
+              rclcpp::Duration::from_nanoseconds(std::numeric_limits<std::int64_t>::max()));
```
ivanpauno, line 207: same comment here

src/domain_bridge/domain_bridge.cpp
ivanpauno: actually, maybe the default value should be `RMW_DURATION_INFINITE` (instead of `-1`) and the `if (lifespan_ns < 0)` is never needed (?)

ivanpauno (approved):
LGTM!! (I've left some minor comments)

After seeing all the parsing code needed I wonder if we should have only enabled qos parameter overrides, though being able to configure everything from the same file looks much nicer!!

src/domain_bridge/domain_bridge.cpp
jacobperron: I wanted an easy way for users to explicitly specify an infinite duration in YAML (hence negative values), though in retrospect we could also use a string `infinite` or `max`.

jacobperron:
I had to add a fix to CI to get a newer version of rclcpp (3951467)

I've cleaned up the commits for a ff-merge.

Issue #14, opened 2021-03-31
Domain bridge crashes with Connext


jacobperron:
I've tried using the bridge with `rmw_connext_cpp` and `rmw_connextdds`, but the program crashes when using either of them.

```
[D0003|Pub(80000008)|T=rt/chatter|CREATE Writer]DDS_DeadlineQosPolicy_is_consistentI:inconsistent QoS policy: period
[D0003|Pub(80000008)|T=rt/chatter|CREATE Writer]DDS_DataWriterQos_is_consistentI:inconsistent QoS policy: deadline
[D0003|Pub(80000008)|T=rt/chatter|CREATE Writer]DDS_Publisher_create_datawriter_disabledI:ERROR: Inconsistent QoS
[ERROR] [1617216731.741343399] [rmw_connextdds]: failed to create DDS writer
[ERROR] [1617216731.741481113] [rmw_connextdds]: failed to allocate RMW_Connext_Publisher
[ERROR] [1617216731.741494351] [rmw_connextdds]: failed to create RMW publisher
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  could not create publisher: Handle's typesupport identifier (rosidl_typesupport_cpp) is not supported by this library, at /home/jacob/ws/ros/latest/src/ros2/rosidl_typesupport/rosidl_typesupport_cpp/src/type_support_dispatch.hpp:111, at /home/jacob/ws/ros/latest/src/ros2/rcl/rcl/src/rcl/publisher.c:114
```

Based on the error message, it doesn't look like Connext is happy with the default deadline value we're providing.

PR #15, opened 2021-04-02
Change default value of deadline and lifespan

jacobperron:
Default to the RMW default value.
This fixes an issue with Connext, which does not handle the same maximum
value as other RMWs.

Fixes https://github.com/ros2/domain_bridge/issues/14

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -126,15 +126,16 @@ The solution is to have the bridge wait until a publisher becomes available befo
 #### Deadline
 
 Due to possible delays induced by the bridge, it is possible that the original deadline of a publisher may not be met when forwarding to a different domain.
-Furthermore, if deadline is important for the user application, then it is probably better for them to set manually on the bridge.
-Therefore, by default the bridge will assume the RMW defined default value for deadline (typically implying "no deadline"), but allow users to opt-in to matching the publisher value or setting the value manually.
+If a deadline is important for the user application, then it should be set manually on the bridge.
+Therefore, by default, the bridge will use the RMW-defined default value for deadline (typically implying "no deadline").
+Users may set the value manually or configure the bridge to use the original publishers value.
 
 #### Lifespan
 
 The lifespan policy is closely coupled with the history policy.
 Since the bridge is not able to automatically detect the history policy of publishers, it is safer to let the user configure lifespan given their application-specific knowledge.
-For similar reasons as the deadline policy, by default the bridge will assume the RMW defined default value for the lifespan policy.
-Likewise, users can opt-in to matching the publisher value or setting the value manually.
+For similar reasons as the deadline policy, by default the bridge will use the RMW-defined default value for the lifespan policy.
+Likewise, users can set the value manually or configure the bridge to use the original publishers value.
 
 #### Liveliness
 
--- a/doc/design.md
+++ b/doc/design.md
@@ -127,13 +127,13 @@ The solution is to have the bridge wait until a publisher becomes available befo
 
 Due to possible delays induced by the bridge, it is possible that the original deadline of a publisher may not be met when forwarding to a different domain.
 Furthermore, if deadline is important for the user application, then it is probably better for them to set manually on the bridge.
-Therefore, by default the bridge will assume a maximum value for deadline, but allow users to opt-in to matching the publisher value or setting the value manually.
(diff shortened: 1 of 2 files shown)


doc/design.md
```diff
@@ -127,13 +127,13 @@ The solution is to have the bridge wait until a publisher becomes available befo
 
 Due to possible delays induced by the bridge, it is possible that the original deadline of a publisher may not be met when forwarding to a different domain.
 Furthermore, if deadline is important for the user application, then it is probably better for them to set manually on the bridge.
```
mjeronimo, line 129: I'm not following this sentence:

".. if deadline is important for the user application, then it is probably better for them to set manually on the bridge."

Is "them" referring to the deadline? If so, it seems to me it should be either:

* "... if deadlines are important for the user application, then they should be set manually on the bridge.", or
* "... if a deadline is important for the user application, then it should be set manually on the bridge."

doc/design.md
```diff
@@ -127,13 +127,13 @@ The solution is to have the bridge wait until a publisher becomes available befo
...
-Therefore, by default the bridge will assume a maximum value for deadline, but allow users to opt-in to matching the publisher value or setting the value manually.
+Therefore, by default the bridge will assume the RMW defined default value for deadline (typically implying "no deadline"), but allow users to opt-in to matching the publisher value or setting the value manually.
```
mjeronimo, line 130: "Therefore, by default the bridge" could be: "Therefore, by default, the bridge"
I suggest "RMW-defined" versus "RMW defined" 

Is this sentence saying that if the user doesn't specify a value for the deadline then instead of using the publisher's value it will use the RMW-defined default value instead? This is how I understand it after reading it a couple of times, but I'm not 100% sure I understand it correctly. Perhaps it could be something like:

"Therefore, by default, the bridge will use the RMW-defined default value for deadlines. Users may direct to use the original value used by the publisher or another value instead by setting..."

jacobperron: Yes, that's correct: If no value is provided by the user, then the RMW-defined value is used. I've broken this down into multiple sentences as suggestion (thanks!). Is it more clear?

doc/design.md
jacobperron: Updated in f5dcaf8
Let me know if it reads better now.

doc/design.md
```diff
@@ -126,15 +126,16 @@ The solution is to have the bridge wait until a publisher becomes available befo
 #### Deadline
 
 Due to possible delays induced by the bridge, it is possible that the original deadline of a publisher may not be met when forwarding to a different domain.
-Furthermore, if deadline is important for the user application, then it is probably better for them to set manually on the bridge.
-Therefore, by default the bridge will assume the RMW defined default value for deadline (typically implying "no deadline"), but allow users to opt-in to matching the publisher value or setting the value manually.
+If a deadline is important for the user application, then it should be set manually on the bridge.
+Therefore, by default, the bridge will use the RMW-defined default value for deadline (typically implying "no deadline").
+Users may set the value manually or configure the bridge to use the original publishers value.
```
mjeronimo, line 131: publishers value -> publisher's value

doc/design.md
```diff
@@ -126,15 +126,16 @@ The solution is to have the bridge wait until a publisher becomes available befo
...
 
 #### Lifespan
 
 The lifespan policy is closely coupled with the history policy.
 Since the bridge is not able to automatically detect the history policy of publishers, it is safer to let the user configure lifespan given their application-specific knowledge.
-For similar reasons as the deadline policy, by default the bridge will assume the RMW defined default value for the lifespan policy.
-Likewise, users can opt-in to matching the publisher value or setting the value manually.
+For similar reasons as the deadline policy, by default the bridge will use the RMW-defined default value for the lifespan policy.
+Likewise, users can set the value manually or configure the bridge to use the original publishers value.
```
mjeronimo, line 138: publisher's

mjeronimo (approved):

doc/design.md
mjeronimo: Yes, clear now, thanks.

doc/design.md
mjeronimo: Yes, it's clear now.

doc/design.md
jacobperron: 87ae3c2

doc/design.md
jacobperron: 87ae3c2

PR #19, opened 2021-05-06
Add topic remapping

jacobperron:
Support remapping the name of a bridged topic.

Added a new topic bridge option with YAML support.

### DIFF
--- a/include/domain_bridge/topic_bridge_options.hpp
+++ b/include/domain_bridge/topic_bridge_options.hpp
@@ -16,6 +16,7 @@
 #define DOMAIN_BRIDGE__TOPIC_BRIDGE_OPTIONS_HPP_
 
 #include <memory>
+#include <string>
 
 #include "rclcpp/callback_group.hpp"
 
@@ -35,6 +36,7 @@ public:
    *
    *    - callback_group = nullptr (node's default)
    *    - qos_options = default (see QosOptions for more information)
+   *    - remap_name = "" (no remap)
    */
   DOMAIN_BRIDGE_PUBLIC
   TopicBridgeOptions() = default;
@@ -73,10 +75,25 @@ public:
   TopicBridgeOptions &
   qos_options(const QosOptions & qos_options);
 
+  /// Get remap name.
+  DOMAIN_BRIDGE_PUBLIC
+  std::string
+  remap_name() const;
+
+  /// Set remap name.
+  /**
+   * An empty string implies no remap.
+   */
+  DOMAIN_BRIDGE_PUBLIC
+  TopicBridgeOptions &
+  remap_name(const std::string & remap_name);
+
 private:
   std::shared_ptr<rclcpp::CallbackGroup> callback_group_{nullptr};
 
   QosOptions qos_options_;
+
+  std::string remap_name_;
 };  // class TopicBridgeOptions
 
 }  // namespace domain_bridge
--- a/include/domain_bridge/topic_bridge_options.hpp
+++ b/include/domain_bridge/topic_bridge_options.hpp
@@ -77,7 +77,7 @@ public:
 
   /// Get remap name.
   DOMAIN_BRIDGE_PUBLIC
-  std::string
+  const std::string &
   remap_name() const;
 
   /// Set remap name.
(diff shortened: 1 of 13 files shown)


include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -73,10 +75,25 @@ class TopicBridgeOptions
   TopicBridgeOptions &
   qos_options(const QosOptions & qos_options);
 
+  /// Get remap name.
+  DOMAIN_BRIDGE_PUBLIC
+  std::string
```
ivanpauno, line 80:
```suggestion
  const std::string &
```

?

test/domain_bridge/config/topic_options.yaml
```diff
@@ -14,9 +14,11 @@ topics:
...
       deadline: 123456
       lifespan: auto
+    remap: bar/remapped
   baz:
     type: test_msgs/msg/Empty
     # Trying other values for deadline and lifespan
     qos:
       deadline: auto
       lifespan: -5
+    remap: ''
```
ivanpauno, line 24: the test should still pass if we delete this line, right?

test/domain_bridge/wait_for_publisher.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <thread>
+
+#include "rclcpp/event.hpp"
+#include "rclcpp/node.hpp"
+
+/// Wait for a publisher to be available (or not)
+/*
+ * \return true if the wait was successful, false if a timeout occured.
+ */
+bool wait_for_publisher(
```
ivanpauno, line 29: Functions defined in a header file must be declared `inline`, to avoid breaking ODR.

(though it won't break ODR if we're only including this private header in test files, but just in case I would still add the `inline` specifier)

test/domain_bridge/wait_for_publisher.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+  while (!predicate() &&
+    time_slept < std::chrono::duration_cast<std::chrono::microseconds>(timeout))
+  {
+    rclcpp::Event::SharedPtr graph_event = node->get_graph_event();
+    node->wait_for_graph_change(graph_event, sleep_period);
+    time_slept = std::chrono::duration_cast<std::chrono::microseconds>(
+      std::chrono::steady_clock::now() - start);
+  }
+  return predicate();
```
ivanpauno, line 50:
```suggestion
  do {
    if (predicate()) {
      return true;
    }
    rclcpp::Event::SharedPtr graph_event = node->get_graph_event();
    node->wait_for_graph_change(graph_event, sleep_period);
    time_slept = std::chrono::duration_cast<std::chrono::microseconds>(
      std::chrono::steady_clock::now() - start);
  } while (time_slept < std::chrono::duration_cast<std::chrono::microseconds>(timeout));
  return predicate();
```

There are two changes there:

 - Avoid the strange case where the fuction returns `false` before the timeout expires (`predicate()` can return a different value each time)
 - Call `predicate()` at least one, even when timeout=0

ivanpauno (approved):
LGTM!

I have left some minor comments

test/domain_bridge/config/topic_options.yaml
jacobperron: Correct. I added this to test an explicit empty string versus no YAML entry.

test/domain_bridge/wait_for_publisher.hpp
jacobperron: Good suggestion. Applied it in 60ffdce.

I cribbed this logic from `test_rclcpp`, so we might consider making similar patches there: https://github.com/ros2/system_tests/blob/8fc14b48bbf3f102fa6b19443c90d1841aca3b9a/test_rclcpp/include/test_rclcpp/utils.hpp

test/domain_bridge/wait_for_publisher.hpp
jacobperron: Made inline 60ffdce

include/domain_bridge/topic_bridge_options.hpp
jacobperron: Done 0d83591

jacobperron:
hmm, seems like my QoS matching tests are flaky :confused:

jacobperron:

PR #21, opened 2021-05-19
Override handle_serialized_message.

clalancette:
This should allow it to build against rolling again.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This has been failing in Rolling for a couple of weeks now: https://build.ros2.org/view/Rbin_uF64/job/Rbin_uF64__domain_bridge__ubuntu_focal_amd64__binary/

Note that this change is *not* backwards-compatible with Galactic, so we'll have to branch off for Galactic to take this fix.

### DIFF
--- a/src/domain_bridge/generic_subscription.hpp
+++ b/src/domain_bridge/generic_subscription.hpp
@@ -69,6 +69,11 @@ public:
   void handle_loaned_message(
     void * loaned_message, const rclcpp::MessageInfo & message_info) override;
 
+  void
+  handle_serialized_message(
+    const std::shared_ptr<rclcpp::SerializedMessage> & serialized_message,
+    const rclcpp::MessageInfo & message_info) override;
+
   // Same as return_serialized_message() as the subscription is to serialized_messages only
   void return_message(std::shared_ptr<void> & message) override;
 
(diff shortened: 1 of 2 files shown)


src/domain_bridge/generic_subscription.hpp
```diff
@@ -69,6 +69,11 @@ class GenericSubscription : public rclcpp::SubscriptionBase
   void handle_loaned_message(
     void * loaned_message, const rclcpp::MessageInfo & message_info) override;
 
+  void
+  handle_serialized_message(
+    const std::shared_ptr<rclcpp::SerializedMessage> & serialized_message,
+    const rclcpp::MessageInfo & message_info) override;
```
jacobperron, line 75: If we remove the `override` does that make it backwards compatible with Galactic?

jacobperron:
Thanks for this. I was confused why this was failing on Rolling and didn't take the time to investigate.

src/domain_bridge/generic_subscription.hpp
clalancette: Yep!  That totally works.  I'm going to put a comment here as well explaining why we aren't marking it as override.

jacobperron:
I've noticed a couple times that the `test_qos_matching` tests fail. I've opened a ticket to look into the flakiness: https://github.com/ros2/domain_bridge/issues/22

jacobperron:

jacobperron (approved):

Issue #22, opened 2021-05-19
test_qos_matching flaky


jacobperron:
I've noticed that sometimes the Jenkins PR jobs result in test failures, but retriggering often resolves the issue.

We should figure out if there's a real bug or if the tests can be improved to reduce flakiness.

Example failure: https://build.ros2.org/job/Gpr__domain_bridge__ubuntu_focal_amd64/9/testReport/

jacobperron:
Another instance: https://build.ros2.org/job/Rdev__domain_bridge__ubuntu_focal_amd64/20/

rebecca-butler:
I've done a bit of investigation into this problem. The flake occurs in tests where we have two publishers on the same topic with different QoS settings. The bridged QoS should use the settings that best match all publishers (e.g. the maximum deadline and lifespan values), but sometimes it only uses the values from the first publisher. The tests that are affected are `qos_matches_max_of_duration_policy` and `qos_matches_topic_exists_multiple_publishers`.

The problem seems to be in `get_topic_qos()`. This function iterates over the info for the available endpoints to get the max lifespan and deadline values, but sometimes only 1 endpoint is available at the time when the function is called. This means it can only use the first publisher's values. 

I don't see an immediately obvious solution to this problem. For now, I've added a delay in `get_topic_qos()` to wait for a bit after the first publisher is found. I'm also checking if the number of endpoints remains the same before and after `get_topic_qos()` runs, and if a new endpoint has become available, the function is called again to update the QoS settings. This isn't exactly a robust solution, but I haven't seen the flake happen since adding it, so it might be good enough. If the issue does come up again, we can just comment out the problematic tests until we have a better solution.

rebecca-butler:
I also discovered a different flaky test while working on this (see #46).

PR #25, opened 2021-06-07
Rti qos profiles patch

ivanpauno:
This is a workaround to be able to use a different participant qos profile in each domain.

For example:

```bash
export RMW_IMPLEMENTATION=rmw_connextdds
export NDDS_QOS_PROFILES=/path/to/examples/EXAMPLE_ROS2TEST_QOS_PROFILES.xml
# The domain bridge executable will try to load a qos profile named "ParticipantProfile_{domain_id}" before creating a participant in each domain
export DOMAIN_BRIDGE_CONNEXT_DDS_QOS_PROFILE_PREFIX=ParticipantProfile
ros2 run domain_bridge domain_bridge src/domain_bridge/examples/example_bridge_config.yaml
```

We probably don't want to merge this in the `main` branch.
We can keep this patch in a separate branch, or only merge the callback trick into 'main` and create a new executable using it in a separate package.


ivanpauno:
@jacobperron FYI, if you have a different idea of how to do this that would be great.

ivanpauno:
I did some testing with this, and it seems to work as expected (I disabled SHMEM transport in only some of the domains and checked that it was set correctly).

src/domain_bridge.cpp
```diff
@@ -104,3 +152,69 @@ int main(int argc, char ** argv)
...
+    for (int i = 0; i < DDS_StringSeq_get_length(&qos_libraries); i++) {
+      const char * item = DDS_StringSeq_get(&qos_libraries, i);
+      if (
+        strcmp("BuiltinQosLib", item) != 0 &&
+        strcmp("BuiltinQosLibExp", item) != 0)
+      {
+        if (!qos_profile_library_name) {
+          qos_profile_library_name = item;
+        } else {
+          // More than one QoS profile library was loaded, do not set
```
mallanmba, line 206: would be helpful to print a warning here

mallanmba:
> We probably don't want to merge this in the `main` branch.
> We can keep this patch in a separate branch, or only merge the callback trick into 'main` and create a new executable using it in a separate package.

Would it be reasonable to merge the callback into the main branch and keep an example of how to use it w/ connext in the repo? This might be something someone else out there would want to do at some point.

ivanpauno:
> Would it be reasonable to merge the callback into the main branch and keep an example of how to use it w/ connext in the repo?

:+1: to merge the callback patch in `main`.

I don't mind to also keep the example with the Connext qos profiles patch here.
I would prefer to not build that example though, or to build it only if Connext was found.
In that way, we avoid depending directly on Connext.

@jacobperron what do you think?

src/domain_bridge.cpp
ivanpauno: Done in https://github.com/ros2/domain_bridge/pull/25/commits/33413665a0760b396cba9c5510ea594af8f4c2e4

ivanpauno:
FYI I noticed that the patch in https://github.com/ros2/domain_bridge/pull/25/commits/3e7faca7409dabe12a15e37fafbc5b9e2444b4b3 is not working correctly (it fails to reset the qos participant profile).
I will look for a fix tomorrow.

ivanpauno:
> FYI I noticed that the patch in 3e7faca is not working correctly (it fails to reset the qos participant profile).

I forgot to set when testing `DOMAIN_BRIDGE_CONNEXT_DDS_QOS_PROFILE_PREFIX`, the only problem was that when that variable was not set the last call to `DDS_DomainParticipantFactory_set_default_participant_qos()` failed.
https://github.com/ros2/domain_bridge/pull/25/commits/baa755096b1029ef1912de8550ebf55c1175c7b4 fixes the issue.

ivanpauno:
The CI issue is unrelated, the problem is that the bridge is only discovering one of the publishers (the reliable one), and so it automatically uses a Reliable profile instead of a best effort one.
The current test is racy, I think we can improve that in a new PR.

ivanpauno:
Rebased, now this branch is incompatible with `galactic`, but it will throw an exception that will make that clear (same behavior as in `main`).

The error message is: ["unexpected callback being called"](https://github.com/ros2/domain_bridge/blob/919d753c788e6655841cf087814e7b3242854b81/src/domain_bridge/generic_subscription.cpp#L72).

https://github.com/ros2/domain_bridge/pull/29 is an easy fix to reintroduce compatibility with galactic, https://github.com/ros2/domain_bridge/pull/30 is a refactor that fixes the issue in a more elegant fashion.

jacobperron:
I'm okay with merging the new callback mechanism. For the Connext-specific example, I'd rather not add any dependencies to Connext to this repo. So we could either keep it as an non-compiled example, or move it to it's own package (like `domain_bridge_examples`). I don't have a strong preference.

jacobperron (commented):
It seems like an okay hack; I'm not sure how else we could do it barring an Connext API we could directly call.

ivanpauno:
Rebased with master

ivanpauno:
> I'm okay with merging the new callback mechanism. For the Connext-specific example, I'd rather not add any dependencies to Connext to this repo.

After the last commits, I'm conditionally compiling the domain bridge executable with the custom patch to be able to configure a different domain participant qos for each domain.
I would like to add a note to the docs and better documentation, but I think we can do that in a follow up PR.

ivanpauno:
Force-pushed to resolve more conflicts

src/domain_bridge_rti_qos.cpp
```diff
@@ -0,0 +1,253 @@
...
+# pragma GCC diagnostic ignored "-Wpedantic"
+# pragma GCC diagnostic ignored "-Wunused-parameter"
+# ifdef __clang__
+#  pragma clang diagnostic ignored "-Wdeprecated-register"
+#  pragma clang diagnostic ignored "-Wreturn-type-c-linkage"
+# endif
+#endif
+#include "ndds/ndds_c.h"
+// #include <ndds/ndds_namespace_cpp.h>
+// #include <ndds/ndds_requestreply_cpp.h>
```
jacobperron, line 38: I guess we can remove these unused includes

jacobperron (approved):
LGTM, one minor comment.

The flaky tests are annoying. I'll see if I can make some time to look into the failures (https://github.com/ros2/domain_bridge/issues/22). Maybe as a short-term workaround we can add a retry flag.

src/domain_bridge_rti_qos.cpp
ivanpauno: Removed https://github.com/ros2/domain_bridge/pull/25/commits/772d374de4e7d4742c67ea1218a906d027f11c2d

ivanpauno:
> The flaky tests are annoying. I'll see if I can make some time to look into the failures (#22). Maybe as a short-term workaround we can add a retry flag.

Agreed, I will take a look to that one soon.

ivanpauno:
Going in, thanks for the review @jacobperron !

PR #26, opened 2021-06-08
Add template method to help bridging services

ivanpauno:
https://github.com/ivanpauno/domain_bridge/tree/ivanpauno/bridge-services-fast-test branch has a modified domain_bridge exectuable wich bridges the "add_two_ints" service (see https://github.com/ivanpauno/domain_bridge/commit/d82984cc3bcd09ec65dd3afe180158c7ff080031 for a diff).

There are some issues in rclcpp API that I have worked arround here, but we should improve in rclcpp:

- It's impossible get a request in a callaback and defer to send a response, that should be possible.
- If the service server that it's being bridged is down, the client will collect  a lot of "pending requests" https://github.com/ros2/rclcpp/blob/7d8b2690ff188672bfe3703d8b3c074b87303931/rclcpp/include/rclcpp/client.hpp#L379.
  The ammout of pending requests can grow indefinetely.
  We could improve rclcpp API, for example:

  ```cpp
  client->send_async_request(request, timeout, handle_request_cb, handle_timeout_cb);
  ```
- There are a lot of cosmetic things in the API signatures that could be improved, e.g. there's no need to take a shared future in the async cliend send request callback, etc.


include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,202 @@
...
+      std::to_string(from_domain_id) << " to domain " << std::to_string(to_domain_id) <<
+      ", ignoring" << std::endl;
+    return;
+  }
+
+  rclcpp::Node::SharedPtr from_domain_node = detail::get_node_for_domain(*impl_, from_domain_id);
+  rclcpp::Node::SharedPtr to_domain_node = detail::get_node_for_domain(*impl_, to_domain_id);
+
+  // Create a client for the 'from_domain'
+  auto client = from_domain_node->create_client<ServiceT>(
```
jacobperron, line 166: Isn't this backwards? I would expect requests to be received by the bridge on the "from_domain" and then forwarded to the "to_domain". Maybe elaborating in the design doc (optionally with a diagram) would help clarify.

include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,202 @@
...
+    options.callback_group());
+
+  auto handle_request =
+    [client](
+      std::shared_ptr<detail::HackedService<ServiceT>> me,
+      const std::shared_ptr<rmw_request_id_t> & request_header,
+      const std::shared_ptr<typename ServiceT::Request> & request) -> void
+    {
+      // TODO(ivanpauno): What do we do if there's no service server available?
+      // This is a general problem in ROS 2, we don't have a clean way of communicating a service error ....
```
jacobperron, line 178: IMO, the domain bridge server should mimic the state of the server on the other side of the bridge. E.g. if the target server is not available, then the domain_bridge should not make it's server available. This should help alleviate any issues with client requests queuing up I hope.

include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,202 @@
...
+      auto result = client->async_send_request(
+        request, [me = std::move(me), request_header](typename rclcpp::Client<ServiceT>::SharedFuture future_response)
+      {
+        // TODO(ivanpauno): Why do we get a future instead of the response directly here?
+        // That doesn't seem to make any sense.
+        // See if we can improve rclcpp service API.
```
jacobperron, line 184: I agree. IIRC something similar was copied over to actions, but we later refactored it to remove the future.

include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,202 @@
...
+/// Bridge a service from one domain to another.
+/**
+ * \param service_bridge: Struct containing info about the service to bridge.
+ * \param options: Options for bridging the topic.
+ */
+template<typename ServiceT>
+void
+DomainBridge::bridge_service(
+  const TopicBridge & service_bridge,
+  const TopicBridgeOptions & options)
```
jacobperron, line 134: nitpick: reusing the "topic" bridge types looks odd. It would look more semantically correct if they were named "ServiceBridge" and "ServiceBridgeOptions" respectively, even if they are just aliases to the topic types.

include/domain_bridge/service_bridge_impl.inc
jacobperron: I understand this may complicate the implementation in the domain_bridge as we'd have to defer the creation of the service server. But it seems like a reasonable expectation from the user. However we ultimately decide, we should add a section about this in the design doc.

include/domain_bridge/service_bridge_impl.inc
ivanpauno: I think it depends of how you think about it:

- You're bridging a service server from domain A to domain B.
- You're bridging requests from domain A to domain B.

Maybe we can be more specific:

- original_server_domain_id
- new_server_domain_id

Those aren't great either, so I'm not sure.
@jacobperron feel free to suggest other naming alternatives.

include/domain_bridge/service_bridge_impl.inc
ivanpauno: Agreed, I will fix that.

include/domain_bridge/service_bridge_impl.inc
ivanpauno: If we create the service only after detecting the service in the original domain, we will also need to detect when the service goes down and destroy the bridging server.
And TBH that really doesn't solve the issue:

- Server in domain A is up.
- Bridging server in domain B is created.
- Client in domain B makes request.
- Server in domain A goes down.
  - Case 1:
    - We destroy server in domain B, client in domain B never gets response.
      (this is already an issue in the non-bridged case)
  - Case 2:
    - Another server in domain A goes up before we destroy the server in domain B.
    - Client in domain B will never get a reply and never see the server in domain B going down.
      (this is also an issue in the non-bridged case)

I would like to see a way of detecting server failures from the client side in `rclcpp`, e.g.:

```cpp
  client->send_async_request(request, timeout, handle_request_cb, handle_timeout_cb);
```

There's no race there.
If the client wasn't available or didn't repply before the timeout, the timeout callback will be triggered (maybe that one could be called the "error" callback).

(there's other alternatives to handle errors correctly, but the above one is the most straightforward IMO)

---

We can try to make the code more complex here in the meantime .... it will make the races less likely to happen but it won't fix the real issues.

include/domain_bridge/service_bridge_impl.inc
ivanpauno: Fixed in https://github.com/ros2/domain_bridge/pull/26/commits/113aa815239703dea2652306b4e0fcb885834ca5.

Now I'm imitating

https://github.com/ros2/domain_bridge/blob/113aa815239703dea2652306b4e0fcb885834ca5/include/domain_bridge/domain_bridge.hpp#L120-L126

signature.

I'm still using `TopicBridge` in the implementation, but it's not actually part of the API so I think it doesn't matter for the moment.

---

I've enforced the [rule of zero](https://en.cppreference.com/w/cpp/language/rule_of_three) in the `ServiceBridgeOptions` class, and also in the current `TopicBridgeOptions` and `QosOptions` classes.

ivanpauno: > I'm still using TopicBridge in the implementation

Fixed that in https://github.com/ros2/domain_bridge/pull/26/commits/752d196acab2ac71a9b8618a6a95cc24f59b181d.

include/domain_bridge/service_bridge_impl.inc
jacobperron: I see how this can be confusing. In my mind, I make the analogy between pub/sub and client/service so I had been thinking that we are bridging the *requests* from one domain into another.

However we decide to interpret a service bridge, we should make sure it is clearly documented in the API and in the design doc. I think a simple diagram of a service bridge to include in the design doc would be nice.

ivanpauno: Sounds good, I will add that to the docs.

I checked the `ros1_bridge`, and I think that it's using a similar terminology to the one here (https://github.com/ros2/ros1_bridge/blob/83390800c5ff205625428f148b514008932cfc61/include/ros1_bridge/factory.hpp#L340), so at least we're being consistent.

include/domain_bridge/service_bridge_impl.inc
jacobperron: I'm don't think we should try and handle the race between a service request and a service going down; this is an issue with services in general, not the domain bridge. 

I'm more concerned with the scenario where the bridge creates a service, even though there is no service to forward requests to. Clients are likely to send requests and never get a response. I think it would be better to only create a service in the bridge once it detects the destination service is available. I would then hope that clients wouldn't have to have special knowledge about how the domain bridge works. E.g. most client code goes something like this:

```c++
client->wait_for_service("foo");
client->send_request(...);
```

If we're dealing with the service bridge as it is currently implemented, then the logic doesn't really make sense since the service being "available" doesn't tell the client anything about the true availability. Does that make sense?

include/domain_bridge/service_bridge_impl.inc
ivanpauno: https://github.com/ros2/domain_bridge/pull/26/commits/a1070609a9f9cfa7acb496e1a32fb38e24ed91c3, https://github.com/ros2/domain_bridge/pull/26/commits/f9fddf4ccae91b3a48bb42fe136058847a2c1eff and https://github.com/ros2/domain_bridge/pull/26/commits/40f1198e69749fdc24bed18ab227a613ee0d1639 added docs explaining how to bridge services.

include/domain_bridge/service_bridge_impl.inc
ivanpauno: Yes, I will open a issues in rclcpp to discuss the problems found here.

doc/service_example.gv
```diff
@@ -0,0 +1,22 @@
+digraph {
+    rankdir="LR";
+    service_in -> node_domain_1 [dir="both"] [label="cli"]
```
jacobperron, line 3:
```suggestion
    service_in -> node_domain_1 [dir="both"] [label="client"]
```

So it's not confused with "command line interface".

doc/service_example.gv
```diff
@@ -0,0 +1,22 @@
...
+    node_domain_1 -> rep_1_to_2
+    rep_1_to_2 -> node_domain_2
+    node_domain_2 -> req_2_to_1
+    req_2_to_1 -> node_domain_1
+    node_domain_2 -> service_out [dir="both"] [label="srv"]
```
jacobperron, line 8:
```suggestion
    node_domain_2 -> service_out [dir="both"] [label="service"]
```

include/domain_bridge/domain_bridge.hpp
```diff
@@ -142,8 +156,12 @@ class DomainBridge
 
 private:
   std::unique_ptr<DomainBridgeImpl> impl_;
+  std::unique_ptr<ServiceBridgeImpl> service_bridge_impl_;
```
jacobperron, line 159: What's the reason (or trade-off) for adding a second pimpl versus bundling the implementation into `DomainBridgeImpl`? For example, `DomainBridgeImpl` could itself have a `ServiceBridgeImpl` object.

It seems like it would be better to include `service_bridge_impl.inc` (or just rename to `service_bridge_impl.hpp`?) directly in `domain_bridge.cpp`. Then we don't have to worry about polluting the namespace here.

include/domain_bridge/domain_bridge.hpp
```diff
@@ -133,6 +135,18 @@ class DomainBridge
     const TopicBridge & topic_bridge,
     const TopicBridgeOptions & options = TopicBridgeOptions());
 
+  /// Bridge a service from one domain to another.
+  /**
+   * \param service_bridge: Struct containing info about the service to bridge.
+   * \param options: Options for bridging the topic.
```
jacobperron, line 141: This docblock is outdated. We should make sure to document exactly what the `from_domain_id` and `to_domain_id` mean here (similar to the design doc).

jacobperron (commented):
Thanks for adding a section (and diagram) to the design doc!

doc/service_example.gv
ivanpauno: Fixed in https://github.com/ros2/domain_bridge/pull/26/commits/a549c594120b6c77d35d67b022c29d45ff1069a7

doc/service_example.gv
ivanpauno: See https://github.com/ros2/domain_bridge/pull/26/commits/752bcefd50aafd642c809c7fc9e51bab65c491d9

include/domain_bridge/domain_bridge.hpp
ivanpauno: > It seems like it would be better to include service_bridge_impl.inc directly in domain_bridge.cpp

We cannot do that because this is a template method, so the definition needs to live in a header file.

> or just rename to service_bridge_impl.hpp?

I followed the google c++ style guide [suggestion](https://google.github.io/styleguide/cppguide.html#Self_contained_Headers):

```
Non-header files that are meant for inclusion should end in .inc and be used sparingly.
```

> What's the reason (or trade-off) for adding a second pimpl versus bundling the implementation into DomainBridgeImpl? For example, DomainBridgeImpl could itself have a ServiceBridgeImpl object.

Template methods make pimpl really complicated, as the definition of the method needs to be available in a header file.
If I move the `DomainBridgeImpl` definition to the header, then I break the idea of `pimpl`, because you cannot guarantee ABI stability anymore (when the "impl" definition is available in a header file and some of the methods definitions too, the compiler can start doing inlining and it's hard to guarantee ABI stability).

The only way to avoid this issues is to add some "type erased" methods to `DomainBridgeImpl` and use them from the template method definition. I think I can manage to do that.

include/domain_bridge/domain_bridge.hpp
ivanpauno: Thanks, updated in https://github.com/ros2/domain_bridge/pull/26/commits/dea8e9972a025ed0ec9807b4beee9734a74692ac.

include/domain_bridge/domain_bridge.hpp
jacobperron: ah, I see the issues now. Let's leave it as-is.

doc/service_example.gv
```diff
@@ -0,0 +1,22 @@
+digraph {
```
jacobperron, line 1: We should regenerate the PNG now.

include/domain_bridge/domain_bridge.hpp
ivanpauno: I refactored a bit in https://github.com/ros2/domain_bridge/pull/26/commits/ec4f8862d05a871beab07cfaa8257a8349b977c6, now there's only one `impl_` member.

doc/service_example.gv
ivanpauno: Updated https://github.com/ros2/domain_bridge/pull/26/commits/be8497289da401f3a9e405cc321cd0429cdd1a23

ivanpauno:
Force pushed after rebasing to fix confilcts.

include/domain_bridge/service_bridge_impl.inc
ivanpauno: Done!

I extended the `WaitForQosHandler` class to also be able to wait for services (it's now names `WaitForGraphEvents`).
I've also refactored the waiting logic a bit in https://github.com/ros2/domain_bridge/pull/26/commits/e5a0eae3a550d51d7bc614a8fa49574556db8f85.

Diff of the commits implementing this: https://github.com/ros2/domain_bridge/pull/26/files/a68b1a811621acd317e868f808c0fba2501881fd..158b8941898c02924312f52e34df2615b1ffe10a.

src/domain_bridge/wait_for_graph_events.hpp
```diff
@@ -91,100 +119,57 @@ class WaitForQosHandler
    *   If QoS settings are already available, then the callback is called immediately.
    */
   void register_on_publisher_qos_ready_callback(
-    std::string topic,
-    rclcpp::Node::SharedPtr node,
+    const std::string & topic,
+    const rclcpp::Node::SharedPtr & node,
```
jacobperron, line 123: Just curious, why switch to a node (versus a shared pointer)? I suppose we don't have to worry about lifetime of the node in either case.

jacobperron (approved):
Just a few minor comments, LGTM!

include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,232 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+
+// #include "rclcpp/rclcpp.hpp"
```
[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/domain_bridge/pull/26/commits/32559888c882a825553dead9b488064840e5a7a3

include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,232 @@
...
+  ServiceBridgeMap bridged_services_;
+};
+
+namespace detail
+{
+// service_name, from_domain_id, to_domain_id
+using ServiceBridge = std::tuple<std::string, size_t, size_t>;
+using ServiceBridgeMap = std::map<
+  ServiceBridge,
+  std::pair<std::shared_ptr<rclcpp::ServiceBase>, std::shared_ptr<rclcpp::ClientBase>>>;
```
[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/domain_bridge/pull/26/commits/ce86f60054d7de3a99590ea1c6775b67276ea28e

include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,232 @@
...
+/// An "awesome" hack on top of `rclcpp::Service`.
+/**
+ * This is needed because you cannot currently get a request and defer to send
+ * a response.
+ * To put it in another way, you cannot spin during a service callback.
+ * 
+ * In this case, we need to call another service in our service callbacks, so we need
+ * this beautiful hack :).
+ * 
+ * TODO(ivanpauno): There should be first class support for this in rclcpp.
```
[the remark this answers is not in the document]
ivanpauno: Done, see https://github.com/ros2/domain_bridge/pull/26/commits/ce86f60054d7de3a99590ea1c6775b67276ea28e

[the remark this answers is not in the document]
ivanpauno: I replied to the wrong comment, sorry.
I've opened https://github.com/ros2/rclcpp/issues/1707.

src/domain_bridge/wait_for_graph_events.hpp
ivanpauno: This is a `Node::SharedPtr &`, not just a `Node &`.
Currently the `waiting_threads_` map is retaining part ownership of the `Node` if there wasn't a thread for that node running already.
The recommendation is to pass ` const shared_ptr<T> &` when you might retain ownership, so it matches the pattern.

---

Currently `WaitForGraphEvents` is retaining part ownership of the passed nodes.
It could also only borrow a reference to a node, and in that case external code should ensure that the node lives for long enough.
But that might make the code harder to understand, so the current approach seems fine.

include/domain_bridge/service_bridge_impl.inc
```diff
@@ -0,0 +1,232 @@
...
+        // That doesn't seem to make any sense.
+        // See if we can improve rclcpp service API.
+
+        auto response = future_response.get();
+        me->send_response(*request_header, *response);
+      });
+    };
+
+  // Create a service for the 'to_domain'
+  // The client should be created first so it is available to the service callback
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/domain_bridge/pull/26/commits/5377aaf95a1b812c8d1df52663a1e1ca5feb6ca3.

src/domain_bridge/wait_for_graph_events.hpp
jacobperron: Makes sense, thanks for the explanation!

jacobperron (approved):

src/domain_bridge/wait_for_graph_events.hpp
```diff
@@ -274,22 +260,91 @@ class WaitForQosHandler
...
+                }
+              }
+            }
+
+            // It's only worth continuing if we have callbacks to handle
+            // or we're shutting down
+            t.cv.wait(
+              lock,
+              [&t]
+              {return (t.topic_callback_vec.size() > 0u) || t.shutting_down;});
```
jacobperron, line 334: We should also check `t.service_callback_vec` here. See #42 for a fix.

PR #32, opened 2021-06-25
Domain bridge container

rebecca-butler:
This PR is related to ros2/rclcpp#1702.


jacobperron (commented):
Nice :+1: 

Instead of adding a new executable, what do you think about making the existing executable a "container"? Then we get the existing CLI options without having to maintain two different executables. I can't really think of any downsides.

rebecca-butler:
> Instead of adding a new executable, what do you think about making the existing executable a "container"?
I agree, this makes sense for reducing code duplication.

jacobperron:
The changes lgtm, pending the decision in https://github.com/ros2/rclcpp/pull/1702. We should document this feature (maybe a new section in the README) and some simple regression tests would be good too.

rebecca-butler:
Following the discussion in ros2/rclcpp#1702, this PR now contains a component manager specific to the domain bridge.

ivanpauno (commented):
I've left a comment about why not to create one `Context` per `Node`, that's the only blocker for me.

---

About adding a `SetNodeOptions()` method as suggested [in this comment](https://github.com/ros2/rclcpp/pull/1702#issuecomment-871518694), I think it's okay but it looks a bit strange.
I really think that it would be okay to add a "domain_id" option to the component manager in `rclcpp_components`. If the implementation guarantees only one `Context` per domain it isn't an issue.
But both solutions are okay to me, I don't mind strongly.

src/domain_bridge/component_manager.cpp
```diff
@@ -0,0 +1,64 @@
...
+        extra_argument.get_value<int>() < 0)
+      {
+        throw rclcpp_components::ComponentManagerException(
+                "Extra component argument 'domain_id' must be a positive integer");
+      }
+      std::size_t domain_id = extra_argument.get_value<std::size_t>();
+      auto context = std::make_shared<rclcpp::Context>();
+      rclcpp::InitOptions init_options;
+      init_options.auto_initialize_logging(false).set_domain_id(domain_id);
+      context->init(0, nullptr, init_options);
```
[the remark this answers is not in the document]
jacobperron: I agree. We should keep track of contexts already created for each domain and and reuse them when adding new nodes to the container.

src/domain_bridge/component_manager.cpp
```diff
@@ -0,0 +1,64 @@
...
+    .arguments(remap_rules);
+
+  for (const auto & a : request->extra_arguments) {
+    const rclcpp::Parameter extra_argument = rclcpp::Parameter::from_parameter_msg(a);
+    if (extra_argument.get_name() == "use_intra_process_comms") {
+      if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
+        throw rclcpp_components::ComponentManagerException(
+                "Extra component argument 'use_intra_process_comms' must be a boolean");
+      }
+      options.use_intra_process_comms(extra_argument.get_value<bool>());
```
jacobperron, line 44: Rather than repeating this block, maybe we could call the parents `SetNodeOptions` first and then modify the options with the domain_bridge specific arguments.

src/domain_bridge/component_manager.cpp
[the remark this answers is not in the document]
rebecca-butler: I've added a map in 19ccada so we can keep track of contexts and their associated domain IDs to allow reuse.

include/domain_bridge/component_manager.hpp
```diff
@@ -0,0 +1,48 @@
...
+protected:
+  DOMAIN_BRIDGE_PUBLIC
+  virtual rclcpp::NodeOptions
+  SetNodeOptions(
+    std::vector<rclcpp::Parameter> parameters,
+    std::vector<std::string> remap_rules,
+    const std::shared_ptr<LoadNode::Request> request);
+
+private:
+  std::map<std::size_t, std::shared_ptr<rclcpp::Context>> contexts_;
```
ivanpauno, line 43: nit: you can use a `std::unordered_map` here

src/domain_bridge/component_manager.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+      if (contexts_.find(domain_id) == contexts_.end() ) {
+        // if context does not already exist for given domain id, create one
+        auto context = std::make_shared<rclcpp::Context>();
+        rclcpp::InitOptions init_options;
+        init_options.auto_initialize_logging(false).set_domain_id(domain_id);
+        context->init(0, nullptr, init_options);
+        contexts_[domain_id] = context;
+      }
+      options.context(contexts_[domain_id]);
```
ivanpauno, line 62: nit^N, feel free to ignore:

I have a small preference for using `emplace()` or `try_emplace()` instead of the subscript operator:

```cpp
      auto it = contexts_.find(domain_id);
      if (it == contexts_.end()) {
        // if context does not already exist for given domain id, create one
        auto context = std::make_shared<rclcpp::Context>();
        rclcpp::InitOptions init_options;
        init_options.auto_initialize_logging(false).set_domain_id(domain_id);
        context->init(0, nullptr, init_options);
        auto it_emplaced_pair = contexts_.emplace(domain_id, context);
        it = it_emplaced_pair.first;
      }
      options.context(*it);
```

ivanpauno (approved):

src/domain_bridge/component_manager.cpp
rebecca-butler: Fixed in 4230582

include/domain_bridge/component_manager.hpp
rebecca-butler: Done in d41047b

src/domain_bridge/component_manager.cpp
rebecca-butler: Done in d41047b, although I changed the last line to `options.context(it->second);`

include/domain_bridge/component_manager.hpp
```diff
@@ -0,0 +1,45 @@
...
+{
+/// ComponentManager handles the services to load, unload, and get the list of loaded components.
+class ComponentManager : public rclcpp_components::ComponentManager
+{
+  using rclcpp_components::ComponentManager::ComponentManager;
+
+protected:
+  DOMAIN_BRIDGE_PUBLIC
+  virtual rclcpp::NodeOptions
+  CreateNodeOptions(const std::shared_ptr<LoadNode::Request> request);
```
jacobperron, line 37:
```suggestion
  rclcpp::NodeOptions
  CreateNodeOptions(const std::shared_ptr<LoadNode::Request> request) override;
```

src/domain_bridge/component_manager.cpp
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "domain_bridge/component_manager.hpp"
+
+#include <memory>
+#include <string>
+#include <vector>
+
+using namespace std::placeholders;
```
jacobperron, line 21: I don't think we need to include these headers (or use the `std::placeholders` namespace)
```suggestion
```

include/domain_bridge/component_manager.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#ifndef DOMAIN_BRIDGE__COMPONENT_MANAGER_HPP__
+#define DOMAIN_BRIDGE__COMPONENT_MANAGER_HPP__
+
+#include "rclcpp_components/component_manager.hpp"
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
```
jacobperron, line 23:
```suggestion
#include <unordered_map>
```

jacobperron (commented):
LGTM

Please add a brief section to the README demonstrating how someone could load a node into the container with a given domain ID.

include/domain_bridge/component_manager.hpp
rebecca-butler: Removed in 2e2e4bd

src/domain_bridge/component_manager.cpp
rebecca-butler: Removed in 2e2e4bd

include/domain_bridge/component_manager.hpp
rebecca-butler: Fixed in 2e2e4bd

jacobperron:
We'll have to wait until the connected PR is merged and released before the CI and RPr jobs will pass. I don't think getting the upstream component manager change into Galactic is likely since it changes ABI (I think adding a new virtual function is not guaranteed to be ABI compatible). So, we'll have to diverge from Galactic (e.g. create a new `galactic` branch).

ivanpauno:
> I think adding a new virtual function is not guaranteed to be ABI compatible

:+1:
Adding a new virtual function breaks ABI.

jacobperron:
I've created a `galactic` branch and opened a rosdistro PR to point to it: https://github.com/ros/rosdistro/pull/30128

rebecca-butler:

rebecca-butler:

jacobperron:
I manually retriggered the GitHub workflow too.

We can ignore the Gpr job since we've already updated rosdistro to point to the `galactic` branch.

jacobperron (approved):

jacobperron:
I guess the workflow may not use our testing repo for fetching Debians (or it's caching an older version of the package). I think we can ignore it and rely on the RPr job.

rebecca-butler:
There's an API change in ros2/rclcpp#1716 that affects this PR, so I think we should wait until that lands before merging this.

rebecca-butler:
I rebased and CI looks good now. @jacobperron could you give this one more look before we merge?

ivanpauno (approved):
LGTM!
It would be good to have some tests for this.

rebecca-butler:
@ivanpauno I added a test for this, could you take a look again? Thanks!

ivanpauno (approved):
LGTM!

CMakeLists.txt
```diff
@@ -132,9 +140,28 @@ ament_export_dependencies(
...
+  rclcpp_components_register_node(${PROJECT_NAME}_test_component_lib
+    PLUGIN "domain_bridge::TestComponent"
+    EXECUTABLE test_component)
+
+  install(TARGETS ${PROJECT_NAME}_test_component_lib
+    EXPORT export_${PROJECT_NAME}
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION lib/${PROJECT_NAME}
+  )
```
jacobperron, line 164: Installing the test library and using the same export name as the primary domain_bridge library causes issues with linking in downstream packages. For instance, I ran into a case where the dependency `test_msgs` could not be found when linking against the domain_bridge.

I don't think we need to install the library (or at the very least we can avoid exporting it).
See https://github.com/ros2/domain_bridge/pull/61 for a fix.

PR #33, opened 2021-06-28
Prevent bridging from/to same domain id

rebecca-butler:


jacobperron (approved):
LGTM!

PR #37, opened 2021-07-07
Update CI to target Galactic

jacobperron:
Targeting the `galactic` branch.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -10,8 +10,8 @@ jobs:
     - uses: ros-tooling/setup-ros@v0.2
       with:
         install-connext: true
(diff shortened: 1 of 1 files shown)


rebecca-butler (approved):
LGTM

ivanpauno (approved):
Isn't the Gpr checker enough for this?
Currently, it's running on `main`, but we can update the branch in `rosdistro` and it will run here.

Maybe the same logic applies for `main`, I don't know how much extra checking this workflow is giving us over the Rpr job.

Anyways, I don't mind having both.

jacobperron:
You're probably right, that the GPr job is enough. I'll leave the workflow job for now, but we can probably just remove it later.

PR #39, opened 2021-07-12
Allow bi-directional topic bridging

rebecca-butler:
Closes #38 

Signed-off-by: Rebecca Butler <rebecca@openrobotics.org>

PR #40, opened 2021-07-16
Add reverse option to swap `from` and `to` domain IDs

rebecca-butler:
As a follow-up to #39, this PR adds an option to reverse the _from_ and _to_ domain IDs when bridging a topic.

### DIFF
--- a/include/domain_bridge/topic_bridge_options.hpp
+++ b/include/domain_bridge/topic_bridge_options.hpp
@@ -104,6 +104,7 @@ private:
   std::string remap_name_;
 
   bool bidirectional_{false};
+
   bool reversed_{false};
 };  // class TopicBridgeOptions
 
--- a/include/domain_bridge/topic_bridge_options.hpp
+++ b/include/domain_bridge/topic_bridge_options.hpp
@@ -37,6 +37,7 @@ public:
    *    - callback_group = nullptr (node's default)
    *    - qos_options = default (see QosOptions for more information)
    *    - remap_name = "" (no remap)
+   *    - reversed = false
    */
   DOMAIN_BRIDGE_PUBLIC
   TopicBridgeOptions() = default;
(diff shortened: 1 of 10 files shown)


include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -85,6 +86,16 @@ class TopicBridgeOptions
   TopicBridgeOptions &
   bidirectional(const bool & bidirectional);
 
+  /// Get reversed option.
+  DOMAIN_BRIDGE_PUBLIC
+  const bool &
+  reversed() const;
+
+  /// Set reversed option.
```
jacobperron, line 94: It might be good to elaborate in the documentation here what happens when this is set to `true`.

src/domain_bridge/domain_bridge.cpp
```diff
@@ -294,9 +294,20 @@ class DomainBridgeImpl
         topic_options.remap_name(), options_.name(), "/");
     }
 
+    // Prevent using reversed flag and bidirectional flag at the same time
```
jacobperron, line 297: Though setting bidirectional to `true` subverts the reversed option, is there actually an issue if we create a bridge with both options set? I'm just wondering if we actually need to print an error here.

include/domain_bridge/topic_bridge_options.hpp
rebecca-butler: Documented this and the `bidirectional` option in 7b06684

src/domain_bridge/domain_bridge.cpp
rebecca-butler: The way that I had written this originally, setting both options would cause the topic to bridge in the same direction twice (which is what I was trying to guard against). But you're right, in its current state there isn't really a problem using both options. I've removed this warning in 7b06684.

jacobperron (approved):
LGTM!

Issue #41, opened 2021-07-20
Service bridge tests fail with non-default RMW


jacobperron:
To reproduce, run the service tests with the env variable `RMW_IMPLEMENTATION` set to something other than `rmw_connextdds`, for example `rmw_fastrtps_cpp` or `rmw_connextdds`:

```
RMW_IMPLEMENTATION=rmw_fastrtps_cpp colcon test --event-handlers console_direct+ --ctest-args -R test_domain_bridge_service
```

We see the following test failures:
```
13: Test timeout computed to be: 60
13: -- run_test.py: invoking following command in '/home/jacob/ws/domain_bridge/build/domain_bridge/test':
13:  - /home/jacob/ws/domain_bridge/build/domain_bridge/test/test_domain_bridge_services --gtest_output=xml:/home/jacob/ws/domain_bridge/build/domain_bridge/test_results/domain_bridge/test_domain_bridge_services.gtest.xml
13: Running main() from gmock_main.cc
13: [==========] Running 1 test from 1 test suite.
13: [----------] Global test environment set-up.
13: [----------] 1 test from TestDomainBridgeServices
13: [ RUN      ] TestDomainBridgeServices.bridge_service
13: /home/jacob/ws/domain_bridge/src/domain_bridge/test/domain_bridge/test_domain_bridge_services.cpp:139: Failure
13: Value of: poll_condition([cli]() {return cli->service_is_ready();}, 3s)
13:   Actual: false
13: Expected: true
13: /home/jacob/ws/domain_bridge/src/domain_bridge/test/domain_bridge/test_domain_bridge_services.cpp:145: Failure
13: Value of: poll_condition([&got_request]() {return got_request.load();}, 3s)
13:   Actual: false
13: Expected: true
13: /home/jacob/ws/domain_bridge/src/domain_bridge/test/domain_bridge/test_domain_bridge_services.cpp:146: Failure
13: Expected equality of these values:
13:   future.wait_for(3s)
13:     Which is: 4-byte object <01-00 00-00>
13:   std::future_status::ready
13:     Which is: 4-byte object <00-00 00-00>
13: [  FAILED  ] TestDomainBridgeServices.bridge_service (9045 ms)
13: [----------] 1 test from TestDomainBridgeServices (9045 ms total)
13: 
13: [----------] Global test environment tear-down
13: [==========] 1 test from 1 test suite ran. (9045 ms total)
13: [  PASSED  ] 0 tests.
[...]

PR #42, opened 2021-07-20
Fix bug when waiting for service to be available

jacobperron:
Also check the vector of service callbacks when deciding if we should continue
executing the wait loop in WaitforGraphEvents threads.

Fixes https://github.com/ros2/domain_bridge/issues/41


rebecca-butler (approved):

ivanpauno:
Thanks for the fix!

PR #43, opened 2021-07-20
Run communication-related tests with all available RMWs

jacobperron:
Follow-up to https://github.com/ros2/domain_bridge/issues/41.
Hopefully this will prevent regressions for RMWs that are not the ROS 2 default.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -26,6 +26,7 @@
   <test_depend>launch</test_depend>
   <test_depend>launch_testing</test_depend>
   <test_depend>launch_testing_ament_cmake</test_depend>
+  <test_depend>rmw_implementation_cmake</test_depend>
   <test_depend>test_msgs</test_depend>
 
   <member_of_group>rosidl_interface_packages</member_of_group>
--- a/package.xml
+++ b/package.xml
@@ -29,6 +29,11 @@
   <test_depend>rmw_implementation_cmake</test_depend>
   <test_depend>test_msgs</test_depend>
 
+  <!-- Ensure the following Tier 1 RMW implementations are available for testing -->
+  <test_depend>rmw_connextdds</test_depend>
+  <test_depend>rmw_cyclonedds_cpp</test_depend>
+  <test_depend>rmw_fastrtps_cpp</test_depend>
+
   <member_of_group>rosidl_interface_packages</member_of_group>
 
   <export>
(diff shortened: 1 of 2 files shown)


jacobperron:
For reference, here is the [documentation for `call_for_each_rmw_implementation`](https://github.com/ros2/rmw/blob/dcd8f836e6c2545b39c1540963e175f06625e2f0/rmw_implementation_cmake/cmake/call_for_each_rmw_implementation.cmake).

jacobperron:
The service tests should fail for `rmw_fastrtps_cpp` and `rmw_connextdds`. However, the Rpr job only installs the default RMW (rmw_cyclonedds_cpp) so we don't see any failures. The GitHub workflow fails since the `action-ros-ci` installs `rmw_fastrtps_cpp` (in addition to cyclonedds). `action-ros-ci` also has an option for installing Connext, which we could enable.

I'm not sure if there's a recommended practice for forcing multiple RMWs to be installed for the buildfarm jobs; I'll ask around.

jacobperron:
I've added explicit test dependencies on tier 1 RMWs to ensure we're testing against them (427676e).

jacobperron:
Rebased now that https://github.com/ros2/domain_bridge/pull/42 is merged. I expect all tests to pass now :crossed_fingers:

rebecca-butler (approved):
All tests are passing :fire: These changes lgtm!

> action-ros-ci also has an option for installing Connext, which we could enable.

This isn't necessary since the Rolling job should catch Connext failures now, but we could do it anyway just to be cautious.

ivanpauno (commented):
LGTM!

For github actions CI, we might want to use this: https://github.com/ros-tooling/setup-ros#Including-RTI-Connext.

ivanpauno (approved):

jacobperron:
> This isn't necessary since the Rolling job should catch Connext failures now, but we could do it anyway just to be cautious.

I think we can disable the GitHub action CI considering we've got the same coverage on build.ros2.org.

jacobperron:
> I think we can disable the GitHub action CI considering we've got the same coverage on build.ros2.org.

On second thought, I'll just leave it for now since it's not broken and it might be useful for running macOS and Windows CI in the future.

ivanpauno:
> On second thought, I'll just leave it for now since it's not broken and it might be useful for running macOS and Windows CI in the future.

That's a good point!

PR #44, opened 2021-07-22
Try depending on rmw_implementation to get RMWs

jacobperron:
Experimenting with a different test dependency.


jacobperron:
This doesn't appear to work, I guess because rmw_implementation declares build-time dependencies to the RMWs instead of runtime dependencies.

Issue #46, opened 2021-07-26
Flaky end-to-end tests


rebecca-butler:
Some of the tests in `test_domain_bridge_end_to_end.cpp` fail occasionally when running `colcon test`:

```
- domain_bridge.TestDomainBridgeEndToEnd compress_mode
  <<< failure message
    /home/rebecca/ws/domain_bridge_ws/domain_bridge/test/domain_bridge/test_domain_bridge_end_to_end.cpp:216
    Value of: poll_condition([&got_message]() {return got_message.load();}, 3s)
      Actual: false
    Expected: true
  >>>
```

I've seen messages like this for the `compress_mode`, `decompress_mode`, and `remap_topic_name` tests.

jacobperron:
@ivanpauno Any ideas about what's happening here?

jacobperron:
If `sub->get_publisher_count() == 1u` does it guarantee that the subscription will receive the one published message?

ivanpauno:
mmm, maybe we also have to wait until `pub->get_subscription_count()` is the correct number of subscriptions (?).
If both conditions are satisfied I would expect the message to always arrive

PR #47, opened 2021-07-27
Wait for multiple publishers in get_topic_qos()

rebecca-butler:
This is an attempt to fix the flaky tests as described in #22.

### DIFF
--- a/src/domain_bridge/wait_for_graph_events.hpp
+++ b/src/domain_bridge/wait_for_graph_events.hpp
@@ -123,7 +123,7 @@ public:
     const rclcpp::Node::SharedPtr & node,
     std::function<void(const QosMatchInfo)> callback)
   {
-    // If the QoS is already avaiable, trigger the callback immediately
+    // If the QoS is already available, trigger the callback immediately
     auto opt_qos = get_topic_qos(topic, *node);
     if (opt_qos) {
       const QosMatchInfo & qos = opt_qos.value();
@@ -199,6 +199,11 @@ private:
     // If there are no publishers, return an empty optional
     if (num_endpoints < 1u) {
       return {};
+    } else {
+      // If we found at least one publisher, wait for a bit (more publishers may become available)
+      std::this_thread::sleep_for(std::chrono::milliseconds(300));
+      endpoint_info_vec = node.get_publishers_info_by_topic(topic);
+      num_endpoints = endpoint_info_vec.size();
     }
 
     // Initialize QoS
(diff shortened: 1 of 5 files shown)


src/domain_bridge/wait_for_graph_events.hpp
```diff
@@ -257,7 +262,15 @@ class WaitForGraphEvents
     result_qos.qos.deadline(max_deadline);
     result_qos.qos.lifespan(max_lifespan);
 
-    return result_qos;
+    // If the number of endpoints has changed, get QoS again
```
jacobperron, line 265: I would expect the time between here and the first time we query for endpoints above is negligible, so I'm not convinced this block is helping much.

src/domain_bridge/wait_for_graph_events.hpp
```diff
@@ -199,6 +199,11 @@ class WaitForGraphEvents
     // If there are no publishers, return an empty optional
     if (num_endpoints < 1u) {
       return {};
+    } else {
+      // If we found at least one publisher, wait for a bit (more publishers may become available)
+      std::this_thread::sleep_for(std::chrono::milliseconds(300));
```
jacobperron, line 204: I would prefer to see this timeout a configurable option (e.g. maybe part of TopicBridgeOptions).

src/domain_bridge/wait_for_graph_events.hpp
```diff
@@ -199,6 +199,11 @@ class WaitForGraphEvents
...
+      endpoint_info_vec = node.get_publishers_info_by_topic(topic);
```
jacobperron, line 205: Rather than querying for endpoints twice, I guess it's equivalent to first sleep and then query once. I.e. I think it's simpler to (optionally) sleep before calling `get_publishers_info_by_topic` above.

src/domain_bridge/wait_for_graph_events.hpp
rebecca-butler: I've added a `delay` option to `TopicBridgeOptions` in 119cf67

src/domain_bridge/wait_for_graph_events.hpp
rebecca-butler: Fair enough, I've removed this in 119cf67

src/domain_bridge/wait_for_graph_events.hpp
rebecca-butler: Yeah, that does seem a bit simpler. Done in 05b6497

jacobperron (approved):

jacobperron:
I've rebased on `main` to see if the latest changes make CI greener.

jacobperron (approved):

jacobperron:
Looks like the only test failure is the known flake https://github.com/ros2/domain_bridge/issues/46

PR #48, opened 2021-07-27
Use unambiguous call to shared_from_this()

jacobperron:
This fixes a compilation error when building against the latest version of rclcpp:

```
 error: request for member 'shared_from_this' is ambiguous
  116 |     callback_(this->shared_from_this(), request_header, typed_request);
      |               ~~~~~~^~~~~~~~~~~~~~~~
```

PR #49, opened 2021-08-04
Use defered service support in rclcpp

ivanpauno:
Replaces https://github.com/ros2/domain_bridge/pull/48.

Multiple inheritance with `std::enable_shared_from_this` is problematic, see:

- https://stackoverflow.com/questions/16082785/use-of-enable-shared-from-this-with-multiple-inheritance
- https://stackoverflow.com/questions/15549722/double-inheritance-of-enable-shared-from-this
- https://stackoverflow.com/questions/12790859/bad-weak-pointer-when-base-and-derived-class-both-inherit-from-boostenable-sha
- https://stackoverflow.com/questions/14939190/boost-shared-from-this-and-multiple-inheritance

The solutions proposed in those threads cannot directly applied here.

To avoid the issue, I'm using the support for defering responses in services added in rclcpp, avoiding the need of using `std::enable_shared_from_this` all together.
We will need a new branch for rolling though, but I couldn't find a backwards compatible solution.

### DIFF
--- a/include/domain_bridge/service_bridge_impl.inc
+++ b/include/domain_bridge/service_bridge_impl.inc
@@ -54,79 +54,6 @@ get_node_for_domain(DomainBridgeImpl & impl, std::size_t domain_id);
 
 const std::string &
 get_node_name(const DomainBridgeImpl & impl);
-
-/// \internal
-/// An "awesome" hack on top of `rclcpp::Service`.
-/**
- * This is needed because you cannot currently get a request and defer to send
- * a response.
- * To put it in another way, you cannot spin during a service callback.
- * 
- * In this case, we need to call another service in our service callbacks, so we need
- * this beautiful hack :).
- * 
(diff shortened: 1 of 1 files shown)


jacobperron (approved):
LGTM. The `main` branch is currently targeting Rolling. We have a separate `galactic` branch now.

jacobperron:
I think the test failures are known flakes and I guess the workflow is failing because it not using the testing repo for Rolling debs. I would go ahead and merge this.

jacobperron:
See https://github.com/ros2/domain_bridge/pull/50 for switching CI to use the testing repo.

ivanpauno:
> LGTM. The main branch is currently targeting Rolling. We have a separate galactic branch now.

Sounds good! I forgot about that

jacobperron:
I've rebased this after merging #50.

PR #51, opened 2021-08-09
Make some end to end tests more reliable

ivanpauno:
Fixes https://github.com/ros2/domain_bridge/issues/46.

### DIFF
--- a/test/domain_bridge/test_domain_bridge_end_to_end.cpp
+++ b/test/domain_bridge/test_domain_bridge_end_to_end.cpp
@@ -142,7 +142,11 @@ TEST_F(TestDomainBridgeEndToEnd, remap_topic_name)
   bridge.bridge_topic(
     topic_name, "test_msgs/msg/BasicTypes", kDomain1, kDomain2, topic_bridge_options);
 
-  EXPECT_TRUE(poll_condition([sub]() {return sub->get_publisher_count() == 1u;}, 3s));
+  EXPECT_TRUE(poll_condition(
+    [sub, pub]() {
+      return sub->get_publisher_count() == 1u && pub->get_subscription_count() == 1u;
+    },
+    3s));
   pub->publish(test_msgs::msg::BasicTypes{});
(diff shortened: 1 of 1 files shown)


ivanpauno:
I tried this patch locally, and it seem that it can sometimes still fail.
I think the issue is that maybe the internal pub/sub that the domain_bridge has created haven't matched the external sub/pub yet.

We can merge this and check if it avoids the issue in CI, if not we will have to add some retrying logic.
We could also use publishers with TRANSIENT_LOCAL durability in the tests, in that case this shouldn't be a problem.

rebecca-butler:
> We could also use publishers with TRANSIENT_LOCAL durability in the tests, in that case this shouldn't be a problem.

It seems like we still have the CI issue with the current patch, so I would support this solution.

jacobperron:
+1 for using transient_local durability in the tests.

ivanpauno:
Done, I tested locally and it seems to be working pretty reliably now.

test/domain_bridge/test_domain_bridge_end_to_end.cpp
```diff
@@ -142,7 +146,6 @@ TEST_F(TestDomainBridgeEndToEnd, remap_topic_name)
   bridge.bridge_topic(
     topic_name, "test_msgs/msg/BasicTypes", kDomain1, kDomain2, topic_bridge_options);
 
-  EXPECT_TRUE(poll_condition([sub]() {return sub->get_publisher_count() == 1u;}, 3s));
```
jacobperron, line 145: I guess waiting for the match is still good to do :shrug:

jacobperron: We could either add a retry (if you think it will help) or skip these tests for rmw_connextdds.

ivanpauno: I did some extra testing locally, it seems to be that connext is just too slow and 3s is sometimes not enough.

jacobperron (approved):

PR #52, opened 2021-08-23
Add option to wait for subscription before bridging a topic

ivanpauno:
Maybe we could first wait for a subscription, and when that's ready register a callback to wait for the publisher qos (?).
Currently I'm not doing that, and QOS settings must by provided through a yaml file (or they will be a default reliable profile).

I have done some manual testing, but after a first review I'm planning to add some automated tests.


jacobperron (commented):
> Maybe we could first wait for a subscription, and when that's ready register a callback to wait for the publisher qos (?).

I guess it depends on how the application nodes are behaving. E.g. if a publisher to the bridge has logic to wait for a matching subscription, then we probably want to be able to create the bridge without waiting on a publisher.  Maybe what we want are two options for the most flexibility:

- `wait_for_publisher`, require that a publisher is available before creating a bridge (defaults to `true`).
- `wait_for_subscription`, require that a subscription is available before creating a bridge (defaults to `false`).

Then, for example, if both options are `true` the bridge will wait for both a pub and sub to be available before creating the bridge. The default values I proposed keep the current behavior.  @ivanpauno What do you think?

ivanpauno:
> @ivanpauno What do you think? 

sounds good to me

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -52,6 +52,12 @@ print_help()
     "                             This overrides any mode set in the YAML file." << std::endl <<
     "                             Defaults to '" << kNormalModeStr << "'." <<
     std::endl <<
+    "    --wait-for-subscription  Will wait for an available subscription before"
+    " bridging a topic" <<
+    std::endl <<
+    "                             This overrides any mode set in the YAML file." << std::endl <<
+    "                             Defaults to '" << kNormalModeStr << "'." <<
```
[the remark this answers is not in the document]
ivanpauno: Thanks, I'm not sure what happened there.
Fixed in https://github.com/ros2/domain_bridge/pull/52/commits/56ea81dd71983de751a51e0f3e9374eefbe6e2b9.

(I have also rebased, but that's the only commit that introduced changes in the last force-push)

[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/domain_bridge/pull/52/commits/a683382feb1fbea67e9b1d74e854a2ceb8b686b4.
Let me know if that matches your idea.

TODO: add tests

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -52,6 +52,12 @@ print_help()
     "                             This overrides any mode set in the YAML file." << std::endl <<
     "                             Defaults to '" << kNormalModeStr << "'." <<
     std::endl <<
+    "    --wait-for-subscription  Will wait for an available subscription before"
+    " bridging a topic." <<
+    std::endl <<
+    "    --dont-wait-for-publisher  Will not wait for an available subscription before"
```
jacobperron, line 58: A few things here:

- I wonder if we really need to add CLI overrides for these options. @mallanmba, are these useful for you?
- nitpick: it's odd to me to see the contraction "don't" in an option. I would prefer "--do-not" or "--no".
- This option is not implemented. 
-  We should clearly document that the CLI args are overriding any values in the YAML config.
- In case the defaults change in the future, it would be nice to have a way to invert the behavior on the CLI. E.g. have mutually exclusive options: `--wait-for-publisher` and `--no-wait-for-publisher`, or a single option that accepts a bool `--wait-for-publisher true|false`

```suggestion
    "    --do-not-wait-for-publisher  Will not wait for an available publisher before"
```

src/domain_bridge/domain_bridge.cpp
```diff
@@ -412,9 +421,41 @@ class DomainBridgeImpl
...
+      case 1:  // WAIT FOR ONLY PUBLISHER, this is the default
+        wait_for_graph_events_.set_delay(topic_options.delay());
+        wait_for_graph_events_.register_on_publisher_qos_ready_callback(
+          topic, from_domain_node, create_bridge);
+        break;
+      case 2:  // WAIT FOR ONLY SUBSCRIPTION
+        wait_for_graph_events_.set_delay(topic_options.delay());
+        wait_for_graph_events_.register_on_subscription_ready_callback(
+          topic, from_domain_node, [create_bridge]() {
+            domain_bridge::QosMatchInfo qos_match;  // with default qos
```
jacobperron, line 442: Are there downsides to matching the subscription QoS?

include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -107,6 +109,34 @@ class TopicBridgeOptions
   TopicBridgeOptions &
   delay(const std::chrono::milliseconds & delay);
 
+  /// Set wait_for_subscription to `value`.
+  /**
+   * When set to true, the domain bridge will wait for a subscription to be available in the
+   * "to domain" before creating the bridge.
```
jacobperron, line 115: I think we should document the interplay between `wait_for_subscription` and `wait_for_publisher` here. E.g. If both are `true`, then the bridge first waits for a subscription and then for a publisher.

We should also describe QoS matching, e.g. if we're waiting on both a subscription and a publisher, then waht QoS will be used for the bridge?

jacobperron (commented):
Good stuff! This is what I had in mind :+1:

include/domain_bridge/process_cmd_line_arguments.hpp
ivanpauno: > I wonder if we really need to add CLI overrides for these options. @mallanmba, are these useful for you?

The `--wait-for-subscription` option is certainly useful.
I'm not sure about the `--no-wait-for-publisher` option.

> nitpick: it's odd to me to see the contraction "don't" in an option. I would prefer "--do-not" or "--no".

done in https://github.com/ros2/domain_bridge/pull/52/commits/f29ee71e2b8980e7ce36ce34586f80d9a5aee356.

> This option is not implemented.

:joy:, done in https://github.com/ros2/domain_bridge/pull/52/commits/f29ee71e2b8980e7ce36ce34586f80d9a5aee356.

> We should clearly document that the CLI args are overriding any values in the YAML config.

I haven't added a YAML config for this options yet, but yes we should document that.

src/domain_bridge/domain_bridge.cpp
ivanpauno: mmm, I had some doubts about that, but after thinking again I think there aren't

include/domain_bridge/process_cmd_line_arguments.hpp
ivanpauno: > In case the defaults change in the future, it would be nice to have a way to invert the behavior on the CLI. E.g. have mutually exclusive options: --wait-for-publisher and --no-wait-for-publisher, or a single option that accepts a bool --wait-for-publisher true|false

Applied the second alternative in https://github.com/ros2/domain_bridge/pull/52/commits/2469cdccc5fb5c4c416ff3fa97be3742c1daba6a

src/domain_bridge/domain_bridge.cpp
ivanpauno: Done in https://github.com/ros2/domain_bridge/pull/52/commits/ec986f36d7ce6de65962a51af83ef8e9b1db8e04.

include/domain_bridge/topic_bridge_options.hpp
ivanpauno: Done in https://github.com/ros2/domain_bridge/pull/52/commits/19835d8811fe87d04d6e5d469b9d8e8ba6900519

include/domain_bridge/process_cmd_line_arguments.hpp
ivanpauno: Fixed a bug in https://github.com/ros2/domain_bridge/pull/52/commits/f792f5dfd3fb02465379897277475fb9610687d0

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -52,6 +52,12 @@ print_help()
     "                             This overrides any mode set in the YAML file." << std::endl <<
     "                             Defaults to '" << kNormalModeStr << "'." <<
     std::endl <<
+    "    --wait-for-subscription true|false  Will wait for an available subscription before"
+    " bridging a topic. Default is false, unless overridden in yaml file or here." <<
```
jacobperron, line 56: nitpick: for consistency with other argument helps strings
```suggestion
    " bridging a topic. This overrides any value set in the YAML file. Defaults to false." <<
```

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -52,6 +52,12 @@ print_help()
...
+    std::endl <<
+    "    --wait-for-publisher true|false  Will wait for an available subscription before"
+    " bridging a topic. Default is true, unless overridden in yaml file or here." <<
```
jacobperron, line 59:
```suggestion
    " bridging a topic. This overrides any value set in the YAML file. Defaults to true." <<
```

jacobperron (commented):
LGTM, just a couple nitpicks.

It would be good to add some tests for the new options before merging.

doc/design.md
```diff
@@ -179,6 +179,17 @@ For example, in the following scenario, messages coming from publisher *B* will
...
+### Waiting for subscription or publisher to bridge a topic
+
+By default, to bridge a topic we wait for an available publisher, and use its qos settings to create the bridge.
+This can be modified by passing the command line arguments:
+
+- `--wait-for-subscription true|false`, default false.
+- `--wait-for-publisher true|false`, default true.
+
+If both are true, first the domain bridge will wait for a publisher and then for a subscription, and use the publisher qos to create the bridge.
+If only wait for subscription is enabled, the qos settings of the subscription found will be used.
```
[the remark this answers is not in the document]
ivanpauno: Done, see https://github.com/ros2/domain_bridge/pull/52/commits/2868b49b2cfd644d16c3a8fd314387cd23585bf5

include/domain_bridge/process_cmd_line_arguments.hpp
ivanpauno: Fixed in https://github.com/ros2/domain_bridge/pull/52/commits/ad1dd930448c2b491a14cba6aaa70f2c6534b3a5

jacobperron (approved):

Issue #59, opened 2021-10-22
Foxy Support


edgarcamilocamacho:
Hello. Should the package work on ROS2 Foxy? When I try to manually build it I get some CMake related errors, and I don't see any  foxy related branch.

jacobperron:
This package was initially developed with ROS Galactic in mind, and I'm not sure if it ever successfully built for Foxy.

I'm not sure I have the time to investigate support for Foxy, but if someone would like to try I suggest starting from the `galactic` branch and trying to adapt it to work with Foxy.

PR #30, opened 2021-06-16
Refactor to use generic pub/sub from rclcpp

rebecca-butler:
Closes #28

PR #60, opened 2021-10-28
Add Foxy support

aprotyas:
### PR description

This PR attempts to make `domain_bridge` a Foxy-compatible package - based off of the `galactic` branch.

The following changes have been introduced:
- Re-add the `GenericPublisher`/`GenericSubscription` classes to this package's source tree. https://github.com/ros2/domain_bridge/pull/60/commits/0c05dab08241af7e4c2695a600e323b1eefd1166
- Cherry-pick some of the compress/decompress mode logic before #24 was rebased on #30. https://github.com/ros2/domain_bridge/pull/60/commits/0c05dab08241af7e4c2695a600e323b1eefd1166
- Re-add `rosidl_target_interfaces` in the CMakeLists.txt file. https://github.com/ros2/domain_bridge/pull/60/commits/0c05dab08241af7e4c2695a600e323b1eefd1166
- Update the `rmw_connextdds` test dependency to read `rmw_connext_cpp`; this package is named differently in Foxy. https://github.com/ros2/domain_bridge/pull/60/commits/0c05dab08241af7e4c2695a600e323b1eefd1166
- Update CI to target Foxy. https://github.com/ros2/domain_bridge/pull/60/commits/cce4273e91f729a98f799ab9aa973626fd8a8a73
- Use `explicit rclcpp::Duration(int64_t ns)` constructor over `rclcpp::Duration::from_nanoseconds`. https://github.com/ros2/domain_bridge/pull/60/commits/9f3b66adcaf892c334330152bf3c174bc3c6d0a1
- Obtain domain ID from `rcl_node_options_t`. https://github.com/ros2/domain_bridge/pull/60/commits/925a5e1798f21c53103e87d4f96f09adc1c9f940
- Use `rmw` types for QoS options class. https://github.com/ros2/domain_bridge/pull/60/commits/699ab3a76559f75a6d8ee30bd40d6951fe32aa2a
- Define a function to convert from `rmw_time_t` to `rclcpp::Duration`. https://github.com/ros2/domain_bridge/pull/60/commits/1562d57950f6bf67a15bc1c42734008783d89a19
- Remove unused "unreachable logic" code. https://github.com/ros2/domain_bridge/pull/60/commits/cfb36cf38a2f9a5f26708323736430c81cc0cbdd
- Update "Galactic" references to read "Foxy" as appropriate. https://github.com/ros2/domain_bridge/pull/60/commits/e614643ec8a3ff3ec4858504c21e1b4e080f64be
- Replace `set_domain_id` functionality by doing the same through the `rcl` layer. https://github.com/ros2/domain_bridge/pull/60/commits/3272a4800db648d4351d287ce5ec85acf5e72701, https://github.com/ros2/domain_bridge/pull/60/commits/549654aaf6047bb28a474d7ffccdbebf0a9886d5, https://github.com/ros2/domain_bridge/pull/60/commits/eb58092084f524447d3c4424bf692ad0b75f1866

### TODOs
I'll just list the things that are missing in Foxy for now:
- ~~Either backport the node option generating hooks added to `rclcpp_components::ComponentManager` in https://github.com/ros2/rclcpp/pull/1702 to Foxy - which is a difficult ask because this is most likely an ABI breaking change, or "revert" https://github.com/ros2/domain_bridge/pull/32, but that may have performance implications because (quoting @rebecca-butler) `"This would allow the bridge and all other nodes to be run in the same process, ideally reducing latency."`~~  My bad, I started from the `main` branch instead of `galactic`.
- [x] `rclcpp::*Policy` enum classes and the respective getters.
These can be accessed through [`rmw/types.h`](https://github.com/ros2/rmw/blob/foxy/rmw/include/rmw/types.h) but I suspect the PR these were introduced in (https://github.com/ros2/rclcpp/pull/1467) can be backported.
  - This is doable without backporting https://github.com/ros2/rclcpp/pull/1467 too, just kind of a slog.
  - https://github.com/ros2/domain_bridge/pull/60/commits/699ab3a76559f75a6d8ee30bd40d6951fe32aa2a.
  - https://github.com/ros2/domain_bridge/pull/60/commits/1562d57950f6bf67a15bc1c42734008783d89a19.
- [x] Domain ID getters/setters.
Technically this functionality is exposed in `rcl_node_options_t`, so it's just a deficiency in the `rclcpp` API (in Foxy). In case `rclcpp::InitOptions::set_domain_id` cannot be backported to Foxy, the `domain_id` can be obtained through `get_rcl_node_options` (minus some mutex concerns..). Not sure how to set the `domain_id` though. Refer to https://github.com/ros2/rclcpp/issues/910 and the PRs that fell out of it.
  - [x] `rclcpp::Context::get_domain_id`.
    - https://github.com/ros2/domain_bridge/pull/60/commits/925a5e1798f21c53103e87d4f96f09adc1c9f940.
  - [x] `rclcpp::InitOptions::set_domain_id`.
    - This can not be replaced using `get_rcl_node_options` since that is a `const` function. Not sure how to compensate for this API without backporting at least https://github.com/ros2/rcl/pull/678, and ideally also https://github.com/ros2/rclcpp/pull/1165 - although this one is most likely not an ABI-compatible patch.
    - My attempt with https://github.com/ros2/domain_bridge/pull/60/commits/3272a4800db648d4351d287ce5ec85acf5e72701, https://github.com/ros2/domain_bridge/pull/60/commits/549654aaf6047bb28a474d7ffccdbebf0a9886d5, https://github.com/ros2/domain_bridge/pull/60/commits/eb58092084f524447d3c4424bf692ad0b75f1866 fails spectacularly, since (of course) modifying a `const` object through `const_cast` is undefined behavior.
- [x] `rclcpp::Duration::from_nanoseconds`.
Use the [`Duration(rcl_duration_value_t nanoseconds)`](https://github.com/ros2/rclcpp/blob/d7804e1b3fd9676d302ec72f02c49ba04cbed5e6/rclcpp/include/rclcpp/duration.hpp#L41-L42) constructor instead.
  - https://github.com/ros2/domain_bridge/pull/60/commits/9f3b66adcaf892c334330152bf3c174bc3c6d0a1.
- [x] Remove stray comments, debug code, etc.
- [x] Document utility functions and other enums/structs (gen pub-sub) introduced in this PR.

Closes https://github.com/ros2/domain_bridge/issues/59.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/test/domain_bridge/test_qos_matching.cpp
+++ b/test/domain_bridge/test_qos_matching.cpp
@@ -58,12 +58,6 @@ std::shared_ptr<rclcpp::Context> TestDomainBridgeQosMatching::context_2_;
 rclcpp::NodeOptions TestDomainBridgeQosMatching::node_options_1_;
 rclcpp::NodeOptions TestDomainBridgeQosMatching::node_options_2_;
 
-std::ostream & operator<<(std::ostream & os, const rclcpp::TopicEndpointInfo & t)
-{
-  os << t.node_namespace() << '/' << t.node_name() << ": " << t.topic_type();
-  return os;
-}
-
 TEST_F(TestDomainBridgeQosMatching, qos_matches_topic_exists_before_bridge)
 {
   const std::string topic_name("test_topic_exists_before_bridge");


test/domain_bridge/test_qos_matching.cpp
```diff
@@ -38,49 +40,67 @@ class TestDomainBridgeQosMatching : public ::testing::Test
 constexpr std::size_t TestDomainBridgeQosMatching::domain_1_;
 constexpr std::size_t TestDomainBridgeQosMatching::domain_2_;
 
+std::ostream& operator<<(std::ostream& os, const rclcpp::TopicEndpointInfo& t)
+{
+  os << t.node_namespace() << '/' << t.node_name() << ": " << t.topic_type();
+  return os;
+}
+
```
aprotyas: This is some code I'm using for debugging the tests. Will remove in a later commit.

test/domain_bridge/test_qos_matching.cpp
```diff
@@ -38,49 +40,67 @@ class TestDomainBridgeQosMatching : public ::testing::Test
...
   // Assert the QoS of the bridged publisher matches
   std::vector<rclcpp::TopicEndpointInfo> endpoint_info_vec =
     node_2->get_publishers_info_by_topic(topic_name);
-  ASSERT_EQ(endpoint_info_vec.size(), 1u);
+  for (auto & t : endpoint_info_vec) {
+    std::cout << t << '\n';
+  }
+  EXPECT_EQ(domain_bridge::utils::get_domain_id_from_node(*node_1), domain_1_);
+  EXPECT_EQ(domain_bridge::utils::get_domain_id_from_node(*node_2), domain_2_);
+  EXPECT_EQ(endpoint_info_vec.size(), 1u);
```
aprotyas, line 81: Same here.

aprotyas (commented):
Just pointing out the fact that I added some code for debugging purposes. Will remove in a later commit.

src/domain_bridge/utils.cpp
```diff
@@ -0,0 +1,103 @@
...
+  .use_global_arguments(false)
+  .start_parameter_services(false)
+  .start_parameter_event_publisher(false);
+
+  auto node = std::make_shared<rclcpp::Node>(name, node_options);
+  auto node_base_interface = node->get_node_base_interface();
+  rcl_node_t * rcl_node_handle = node_base_interface->get_rcl_node_handle();
+  // Hacky work-around because setting domain ID is not a feature in the rclcpp layer
+  const_cast<rcl_node_options_t *>(
+    rcl_node_get_options(rcl_node_handle))->domain_id = domain_id;
```
jacobperron, line 74: IIRC, since Foxy we no longer have a 1-to-1 mapping from ROS nodes to DDS participants. Instead, a participant is mapped to each ROS context. Therefore, the domain ID is tied to the ROS context, not the node. I suspect the problem is that the domain ID being set in the node options are not being applied.

Unfortunately, it doesn't look like an `rcl` API exists in Foxy to set the domain ID. I think what we want to do is set this member of the RMW struct `rmw_init_options`:

https://github.com/ros2/rmw/blob/6d23f9fc6bfe4616e1efeeefee7a927635a83afe/rmw/include/rmw/init_options.h#L52

I suppose one possible fix is to backport the missing `rcl` setter:

https://github.com/ros2/rcl/blame/master/rcl/include/rcl/init_options.h#L174

and set it here before creating the context (ie. backport https://github.com/ros2/rcl/pull/678).

---

Or, we could try setting the environment variable `ROS_DOMAIN_ID` before creating the ROS context. This is what eProsima's Integration Service does (or used to): https://github.com/eProsima/Integration-Service/blob/a76ccb63fbdb9f8193fd252c6c8d1277c5f1e731/packages/ros2/src/SystemHandle.cpp#L142-L149

Note, it also restores the value of `ROS_DOMAIN_ID` when it is done.

aprotyas: I tried both setting/resetting the environment variable `ROS_DOMAIN_ID` and tried initializing a `rclcpp::Context` with a `rclcpp::InitOptions` constructed with a domain ID-set `rcl_init_options_t`, but neither seemed to work. Will have to investigate a little further. The rpr job will fail as expected though because https://github.com/ros2/rcl/pull/946 hasn't been released yet.

antonyramsy:
Hi. I just tried to test this branch with with ROS foxy, but i encountered an error when a node begins to publish some data. I used the `example_bridge_config.yaml` file as mentioned in the readme file and published data in the chatter topic. the error is:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  unexpected callback being called

```

this is my `example_bridge_config.yaml` file:
```
name: my_bridge
from_domain: 0
to_domain: 3
topics:
  chatter:
    type: example_interfaces/msg/String
```
am i missing something alongway doing it or it's known issue?

aprotyas:
@antonyramsy this branch was still WIP and I didn't have enough bandwidth to get back to it for a while now.

I remember seeing these errors when I was testing this branch too, and the problem was that in Foxy, `rcl` offers no such API to set domain IDs -- check https://github.com/ros2/domain_bridge/pull/60#discussion_r742555634 for what I'd already (unsuccessfully) tried. I just never had time to fully verify that my approach in https://github.com/ros2/domain_bridge/pull/60/commits/efa97ae33fed98103a5030cf70e605dc7295ea1a works.

I might have some time to revive this PR soon-ish, but if you need this branch to work I think the way to go would be to run the tests in this package and debug the errors from there. As it is, I think your configuration is correct but the branch is not ready to land.

antonyramsy:
@aprotyas  Thanks a lot for the response. I will consider using this package along with other approaches i can take, and will inform if any progress in the development has taken place.

aprotyas:
@jacobperron as of ff41c31, the end-to-end tests pass in this branch. I'm still struggling against the default lifespan and deadline values in QOS profiles.

From https://github.com/ros2/rmw/pull/301#pullrequestreview-595709837, it looks like in Foxy, the default zero values for the lifespan and deadline means each RMW uses its own default value, which is default to test against compared to the "max" duration enforced post-Foxy (i.e. `rclcpp::Duration(std::numeric_limits<int64_t>::max())`). Is there a way around this?

antonyramsy:
@aprotyas  I've tested this branch. first i got this error:
```
CMake Deprecation Warning at /opt/ros/humble/share/rosidl_cmake/cmake/rosidl_target_interfaces.cmake:32 (message):
  Use rosidl_get_typesupport_target() and target_link_libraries() instead of
  rosidl_target_interfaces()
```

and changed `rosidl_target_interfaces()` with `rosidl_get_typesupport_target()`. after that another error appeared which looks like the message is not being created:
```
Starting >>> domain_bridge
--- stderr: domain_bridge                             
~/domain_bridge_ws/src/src/domain_bridge/domain_bridge.cpp:43:10: fatal error: domain_bridge/msg/compressed_msg.hpp: No such file or directory
   43 | #include "domain_bridge/msg/compressed_msg.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
full text log:
[log.txt](https://github.com/ros2/domain_bridge/files/8813437/log.txt)

antonyramsy:
@aprotyas Hi, sorry for delay. I've tested the updated code both on foxy and humble. in normal condition(sending some data in one domain and receiving it in another one) both if them are ok. but i realized that bidirectional bridging in ROS foxy is not working as i expect (maybe i'm doing it in the wrong way) but in humble it works fine(by using the same YAML config file in different versions of ROS). here is the config i'm using to test bidirectional bridge:
```
name: my_bridge
from_domain: 2
to_domain: 3
topics:
  chatter:
    type: example_interfaces/msg/String
    remap: chitter
    bidirectional: True
```
with this config i was able to publish some date on topic `chatter` from domain 2 and receive it in domain 3 under `chitter` topic. and vice versa(publishing in domain 3 in topic `chatter` and receiving it in domain 2 in topic `chitter` ) in ROS humble.

but doing the same in ROS foxy, although you are able to send data from domain 2 to 3, but you wont be able to see published data from domain 3 in topic `chatter` in domain 2 in topic `chitter` (the topic `chitter` does not show up in `ros2 topic list` command result.)

PR #61, opened 2021-11-05
Do not associate test library with export

jacobperron:
Otherwise, downstream targets will try to link against it.
This can cause issues linking because the test component library is not exporting dependencies (e.g. test_msgs).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,12 @@ if(BUILD_TESTING)
   rclcpp_components_register_node(${PROJECT_NAME}_test_component_lib
     PLUGIN "domain_bridge::TestComponent"
     EXECUTABLE test_component)
+
+  install(TARGETS ${PROJECT_NAME}_test_component_lib
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION lib/${PROJECT_NAME}
+  )
 endif()
 
 ament_package()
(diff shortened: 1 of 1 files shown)


jacobperron:
hmm, looks like this patch doesn't work. I guess it only appeared to work locally because I forgot to clean my build folder.

Seems like there's just not a convenient way to write test components without installing them (probably something we could improve in `rclcpp_components`). For example, `rclcpp_components` works around the issue by not calling `rclcpp_components_register_node` and mocking an ament index:

https://github.com/ros2/rclcpp/blob/e0e96681d9f0f5bd1f220b3d3107a612dad8321b/rclcpp_components/CMakeLists.txt#L83-L111

:slightly_frowning_face:

jacobperron:
Removing the association with the export fixes the issue for me (c86d28e).

CMakeLists.txt
```diff
@@ -158,7 +158,6 @@ if(BUILD_TESTING)
     EXECUTABLE test_component)
 
   install(TARGETS ${PROJECT_NAME}_test_component_lib
-    EXPORT export_${PROJECT_NAME}
```
aprotyas, line 161: I don't know why this was originally exported, but it could also be exported as `export_${PROJECT_NAME}_test` or something to prevent accidental linking against? Although to be fair, there isn't a reason to install (or export) the test library.

ivanpauno: yeah, installing or exporting the library sounds like a bad idea.
The workaround that rclcpp_components is using looks better, though this seems fine to solve the CI issues.

ivanpauno (approved):

jacobperron:
I'm not sure why the GitHub workflow is suddenly using Python 2.7...
With Rpr passing, I'm going to merge this. I'll play around with the GitHub workflow separately.

jacobperron:
IIUC, the GitHub workflow requires some changes to rosidl_parser to become available. It should hopefully resolve itself once a Rolling sync to testing happens: https://build.ros2.org/job/Rrel_sync-packages-to-testing_focal_amd64/

PR #63, opened 2021-11-24
Auto remove bridge when endpoint removed

ivanpauno:
The callbacks added to the `WaitForGraphEvents` object are now kept forever instead of being removed after the first time they are called.
There might be another way to get this working (register callback waiting for publishing/subscription, remove it and register the one waiting for publisher/subscription/removed, ...), though I find that more complex to implement in the current code.

### DIFF
--- a/include/domain_bridge/process_cmd_line_arguments.hpp
+++ b/include/domain_bridge/process_cmd_line_arguments.hpp
@@ -63,8 +63,7 @@ print_help()
     " passed, when either the subscrption or publisher is removed also will the bridge."
     " If only --wait-for-subscription was passed, the bridge will only be removed if the"
     " subscription is."
-    " The bridge will be created again when the original \"wait for\" condition is satisfied"
-    " again." <<
+    " The bridge will be recreated when the original \"wait for\" condition is satisfied." <<
     std::endl <<
     "    --help, -h               Print this help message." << std::endl;
 }
--- a/include/domain_bridge/process_cmd_line_arguments.hpp
+++ b/include/domain_bridge/process_cmd_line_arguments.hpp
@@ -58,8 +58,8 @@ print_help()
     "    --wait-for-publisher true|false  Will wait for an available subscription before"
     " bridging a topic. This overrides any value set in the YAML file. Defaults to true." <<
     std::endl <<
-    "    --auto-remove true|false  If true, the bridge will be removed if endpoint was waited"
-    " for is removed. i.e. if both --wait-for-subscription and --wait-for-publisher where "
+    "    --auto-remove true|false  If true, the bridge will be removed if the endpoint that was waited"
+    " on is removed. I.e. if both --wait-for-subscription and --wait-for-publisher were "
     " passed, when either the subscrption or publisher is removed also will the bridge."
     " If only --wait-for-subscription was passed, the bridge will only be removed if the"
     " subscription is."
--- a/include/domain_bridge/process_cmd_line_arguments.hpp
+++ b/include/domain_bridge/process_cmd_line_arguments.hpp
@@ -60,7 +60,7 @@ print_help()
     std::endl <<
     "    --auto-remove true|false  If true, the bridge will be removed if the endpoint that was waited"
     " on is removed. I.e. if both --wait-for-subscription and --wait-for-publisher were "
-    " passed, when either the subscrption or publisher is removed also will the bridge."
+    " passed, then when either the subscription or publisher is removed, the bridge will also be removed."
     " If only --wait-for-subscription was passed, the bridge will only be removed if the"
     " subscription is."
     " The bridge will be recreated when the original \"wait for\" condition is satisfied." <<
--- a/include/domain_bridge/process_cmd_line_arguments.hpp
+++ b/include/domain_bridge/process_cmd_line_arguments.hpp
@@ -268,7 +268,7 @@ process_cmd_line_arguments(const std::vector<std::string> & args)
   domain_bridge::DomainBridgeConfig domain_bridge_config =
     domain_bridge::parse_domain_bridge_yaml_config(*yaml_config);
 
-  // Override 'from_domain','to_domain' and 'wait_for_subscription' in config
+  // Override domain bridge configuration options
   if (
     from_domain_id || to_domain_id || wait_for_subscription || wait_for_publisher ||
     auto_remove)
--- a/include/domain_bridge/process_cmd_line_arguments.hpp
+++ b/include/domain_bridge/process_cmd_line_arguments.hpp
@@ -58,9 +58,10 @@ print_help()
     "    --wait-for-publisher true|false  Will wait for an available subscription before"
     " bridging a topic. This overrides any value set in the YAML file. Defaults to true." <<
     std::endl <<
-    "    --auto-remove true|false  If true, the bridge will be removed if the endpoint that was waited"
-    " on is removed. I.e. if both --wait-for-subscription and --wait-for-publisher were "
-    " passed, then when either the subscription or publisher is removed, the bridge will also be removed."
+    "    --auto-remove true|false  If true, the bridge will be removed if the endpoint that was"
+    " waited on is removed. I.e. if both --wait-for-subscription and --wait-for-publisher were"
+    " passed, then when either the subscription or publisher is removed, the bridge will also be"
+    " removed."
     " If only --wait-for-subscription was passed, the bridge will only be removed if the"
     " subscription is."
     " The bridge will be recreated when the original \"wait for\" condition is satisfied." <<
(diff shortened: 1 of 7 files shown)


include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -141,6 +142,35 @@ class TopicBridgeOptions
...
+   *
+   * If set to Autoremove::OnNoSubscription the domain bridge will stop bridging when no
+   * subscription is available in the "to domain".
+   * The bridge will automatically be created again when a new subscription is discovered.
+   *
+   * When set to Autoremove::Disabled, the bridge will keep running for ever once created.
+   */
+  DOMAIN_BRIDGE_PUBLIC
+  TopicBridgeOptions &
+  auto_remove(AutoRemove value);
```
jacobperron, line 167: I wonder if having an enum here instead of a boolean is exposing _too_ much configurability. Having a scenario where the "wait for" options are not symmetric with the auto-removal policy seems odd to me, but maybe there's a use-case.

I suppose either way it doesn't really affect the complexity of the implementation, so we can leave it as-is. I was just curious if you could think of a scenario where we want this level of configuration.

src/domain_bridge/wait_for_graph_events.hpp
```diff
@@ -150,7 +180,10 @@ class WaitForGraphEvents
   struct TopicAndCallback
   {
     std::string topic;
-    std::function<void(const QosMatchInfo)> cb;
+    std::variant<
```
jacobperron, line 183: `#include <variant>`

package.xml
```diff
@@ -28,6 +28,7 @@
   <test_depend>launch_testing</test_depend>
   <test_depend>launch_testing_ament_cmake</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
+  <test_depend>rosgraph_msgs</test_depend>
```
jacobperron, line 31: remove? or maybe part of future tests?

src/domain_bridge/domain_bridge.cpp
```diff
@@ -453,6 +459,30 @@ class DomainBridgeImpl
...
     }
+    if (
+      domain_bridge::TopicBridgeOptions::AutoRemove::OnNoPublisher ==
+      topic_options.auto_remove() ||
+      domain_bridge::TopicBridgeOptions::AutoRemove::OnNoPublisherOrSubscription ==
+      topic_options.auto_remove())
+    {
+      this->wait_for_graph_events_.register_on_no_publisher_callback(
+        topic, from_domain_node, [this, topic_bridge]() {
+          this->bridged_topics_[topic_bridge] = {nullptr, nullptr};
```
jacobperron, line 470: It looks like there's a data race here (and inside `create_bridge` above, in retrospect). Should we put a lock guard on `bridged_topics_`?

jacobperron (commented):
Looks like a reasonable approach to me. We should add some tests to exercise the new functionality. WaitForGraphEvents is getting complex :sweat_smile: 

More comments below.

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -58,6 +58,14 @@ print_help()
     "    --wait-for-publisher true|false  Will wait for an available subscription before"
     " bridging a topic. This overrides any value set in the YAML file. Defaults to true." <<
     std::endl <<
+    "    --auto-remove true|false  If true, the bridge will be removed if endpoint was waited"
+    " for is removed. i.e. if both --wait-for-subscription and --wait-for-publisher where "
```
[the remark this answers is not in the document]
ivanpauno: thanks!

include/domain_bridge/topic_bridge_options.hpp
ivanpauno: > Having a scenario where the "wait for" options are not symmetric with the auto-removal policy seems odd to me

Yes, I agree.

> I suppose either way it doesn't really affect the complexity of the implementation, so we can leave it as-is.

:+1:

> I was just curious if you could think of a scenario where we want this level of configuration.

Not really, an independent option was the first implementation that came to mind.

package.xml
ivanpauno: https://github.com/ros2/domain_bridge/blob/348fd60576b841e2ed5fb24353b9205d738aad99/examples/example_bridge_config.yaml#L8
https://github.com/ros2/domain_bridge/blob/348fd60576b841e2ed5fb24353b9205d738aad99/test/test_domain_bridge.launch.py#L63-L64

I found that the dependency was missing while working with a partially-built ros2 workspace as a base.

include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -141,6 +142,35 @@ class TopicBridgeOptions
...
+  /**
+   * If set to Autoremove::OnNoPublisher the domain bridge will stop bridging when no publisher is
+   * available in the "from domain".
+   * The bridge will automatically be created again when a new publisher is discovered.
+   *
+   * If set to Autoremove::OnNoSubscription the domain bridge will stop bridging when no
+   * subscription is available in the "to domain".
+   * The bridge will automatically be created again when a new subscription is discovered.
+   *
+   * When set to Autoremove::Disabled, the bridge will keep running for ever once created.
```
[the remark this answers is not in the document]
ivanpauno: fixed in https://github.com/ros2/domain_bridge/pull/63/commits/effe9b7b4056d23784c27529aa9b4916511949b7

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -58,6 +58,14 @@ print_help()
...
+    " passed, when either the subscrption or publisher is removed also will the bridge."
+    " If only --wait-for-subscription was passed, the bridge will only be removed if the"
+    " subscription is."
+    " The bridge will be created again when the original \"wait for\" condition is satisfied"
+    " again." <<
```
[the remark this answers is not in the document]
ivanpauno: fixed in https://github.com/ros2/domain_bridge/pull/63/commits/902facbc6256907f06722468d9ed08ad78ef611a

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -58,6 +58,14 @@ print_help()
     "    --wait-for-publisher true|false  Will wait for an available subscription before"
     " bridging a topic. This overrides any value set in the YAML file. Defaults to true." <<
     std::endl <<
+    "    --auto-remove true|false  If true, the bridge will be removed if endpoint was waited"
+    " for is removed. i.e. if both --wait-for-subscription and --wait-for-publisher where "
+    " passed, when either the subscrption or publisher is removed also will the bridge."
```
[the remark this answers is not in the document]
ivanpauno: done, see https://github.com/ros2/domain_bridge/pull/63/commits/9b23afdb57f5c03d82a4f513eee8ab35d712c020

include/domain_bridge/process_cmd_line_arguments.hpp
```diff
@@ -213,7 +270,10 @@ process_cmd_line_arguments(const std::vector<std::string> & args)
     domain_bridge::parse_domain_bridge_yaml_config(*yaml_config);
 
   // Override 'from_domain','to_domain' and 'wait_for_subscription' in config
```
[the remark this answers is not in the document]
ivanpauno: solved in https://github.com/ros2/domain_bridge/pull/63/commits/73d9d3a160d0b7ed622c2f1400a13e5b417606f4

include/domain_bridge/topic_bridge_options.hpp
```diff
@@ -141,6 +142,35 @@ class TopicBridgeOptions
...
+  {
+    OnNoPublisher,
+    OnNoSubscription,
+    OnNoPublisherOrSubscription,
+    Disabled,
+  };
+
+  /// Set auto_remove to value.
+  /**
+   * If set to Autoremove::OnNoPublisher the domain bridge will stop bridging when no publisher is
```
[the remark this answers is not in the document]
ivanpauno: fixed in https://github.com/ros2/domain_bridge/pull/63/commits/19e8e9c691149cc0e1468be42849321e3f683b74

src/domain_bridge/wait_for_graph_events.hpp
ivanpauno: Done in https://github.com/ros2/domain_bridge/pull/63/commits/2eab5ce89664146c259c77e56d7a8f880dc3fc05

src/domain_bridge/domain_bridge.cpp
ivanpauno: Yes, there's a data race here.
I think that adding a mutex is reasonable, I'll take a look.

package.xml
jacobperron: ah, good catch.

ivanpauno:
> We should add some tests to exercise the new functionality.

See https://github.com/ros2/domain_bridge/pull/63/commits/7742dec8d417808d0531d6eabf112b9478222a99

src/domain_bridge/domain_bridge.cpp
ivanpauno: fixed in https://github.com/ros2/domain_bridge/pull/63/commits/0fbe05b99715709866ecaadcbb23caf52ca9cfcb

jacobperron (approved):

PR #66, opened 2022-05-11
Fixes for Ubuntu Jammy

jacobperron:
* Fix -Wvesing-parse compiler warning (1d5dbc8bdd348f8d23cf35aca8fb99f319a52a11)
Use braces to initialize variable instead of parentheses.
* Uncrustify (578137bbe7958a2efc08c95fe7624426fa9fc56b)

### DIFF
--- a/test/domain_bridge/test_domain_bridge.cpp
+++ b/test/domain_bridge/test_domain_bridge.cpp
@@ -47,7 +47,7 @@ TEST_F(TestDomainBridge, construction_destruction)
   }
   // With options
   {
-    domain_bridge::DomainBridge bridge(domain_bridge::DomainBridgeOptions());
+    domain_bridge::DomainBridge bridge{domain_bridge::DomainBridgeOptions()};
   }
 }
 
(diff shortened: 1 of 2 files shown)


christophebedard:
Bumping ros-tooling/setup-ros to v0.3 should fix the error(s).

jacobperron:
@christophebedard Thanks, trying it out now.

jacobperron:
Looks like the setup-ros action is trying to install `rti-connext-dds-5.3.1`, which is not available for Rolling/Humble. Instead we should be installing `rti-connext-dds-6.0.1`.

jacobperron:
I don't think we need to block this PR on the GitHub Action CI. I think the Hpr and Rpr jobs are sufficient.

christophebedard:
> Looks like the setup-ros action is trying to install `rti-connext-dds-5.3.1`, which is not available for Rolling/Humble. Instead we should be installing `rti-connext-dds-6.0.1`.

We somehow missed this and apparently don't have a test for it! I'll fix it. You could also remove `install-connext: true`.

jacobperron:
We have extra (optional) features in this package that depend explicitly on Connext, so it's nice to have it installed if possible.

ivanpauno (approved):
LGTM!

PR #68, opened 2022-06-01
[domain_bridge] Improve doxygen docblock layout

aprotyas:
Prior to this patch, the `\see` tag would render a `See also` section in
the middle of a running sentence, and hence would look out-of-place.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>


jacobperron (approved):
Thanks for the patch :bow:

PR #70, opened 2022-09-29
Fix deprecation warnings

jacobperron:
Pass rclcpp instance instead of rmw instance to create_client and create_service methods.


jacobperron:
Since the new API isn't available in humble, I'll create a new branch in this repo for humble.

jacobperron:
Branch for humble: https://github.com/ros2/domain_bridge/pull/71

PR #73, opened 2022-12-21
Support configuring liveliness in yaml files

ivanpauno:


include/domain_bridge/qos_options.hpp
```diff
@@ -121,20 +121,47 @@ class QosOptions
    */
   DOMAIN_BRIDGE_PUBLIC
   QosOptions &
-  lifespan(const std::int64_t & lifespan);
```
adityapande-1995, line 124: Are ``deadline`` and ``lifespan`` meant to be changed ? Maybe we can leave these as const references then.

include/domain_bridge/qos_options.hpp
```diff
@@ -87,7 +87,7 @@ class QosOptions
    * If nullopt_t is returned, then use automatic matching.
    */
   DOMAIN_BRIDGE_PUBLIC
-  std::optional<std::int64_t>
```
adityapande-1995, line 90: Should we also clean up ``std::`` in the function declarations above ? Like in ``depth(const std::size_t & depth);``

src/domain_bridge/domain_bridge.cpp
```diff
@@ -402,9 +402,23 @@ class DomainBridgeImpl
...
-
-        qos.liveliness(qos_match.qos.liveliness());
-        qos.liveliness_lease_duration(qos_match.qos.liveliness_lease_duration());
+        if (qos_options.liveliness()) {
+          qos.liveliness(*qos_options.liveliness());
+        } else {
+          qos.liveliness(qos_match.qos.liveliness());
+        }
+        if (qos_options.liveliness_lease_duration()) {
+          const auto liveliness_lease_duration_ns = *qos_options.liveliness_lease_duration();
```
adityapande-1995, line 411: nit : Alignment

src/domain_bridge/domain_bridge.cpp
```diff
@@ -402,9 +402,23 @@ class DomainBridgeImpl
...
+          if (liveliness_lease_duration_ns < 0) {
+            qos.liveliness_lease_duration(
+              rclcpp::Duration::from_nanoseconds(std::numeric_limits<std::int64_t>::max()));
```
adityapande-1995, line 414: nit : Alignment

src/domain_bridge/parse_domain_bridge_yaml_config.cpp
```diff
@@ -142,6 +142,34 @@ static QosOptions parse_qos_options(YAML::Node yaml_node, const std::string & fi
...
+      auto liveliness_str = qos_node["liveliness"].as<std::string>();
+      if (0u == std::strcmp(liveliness_str.c_str(), "manual_by_topic")) {
+        options.liveliness(rclcpp::LivelinessPolicy::ManualByTopic);
+      } else if (0u == std::strcmp(liveliness_str.c_str(), "automatic")) {
+        options.liveliness(rclcpp::LivelinessPolicy::Automatic);
+      } else if (0u == std::strcmp(liveliness_str.c_str(), "system_default")) {
+        options.liveliness(rclcpp::LivelinessPolicy::SystemDefault);
+      } else {
+        throw YamlParsingError(
+                file_path, "livelines must be manual_by_topic/automatic/system_default");
```
adityapande-1995, line 156: nit : Alignment

ivanpauno: this is what uncrustify wanted, if not it doesn't pass.
(see that CI is passing)

src/domain_bridge/domain_bridge.cpp
ivanpauno: what do you mean exactly?
This one looks fine, it's one line.

Uncrustify is passing.

src/domain_bridge/domain_bridge.cpp
ivanpauno: same, this one looks fine to me

include/domain_bridge/qos_options.hpp
ivanpauno: There's no reason to pass integers as `const &`.
I think it's a good idea to change it now, it shouldn't affect domain bridge users.

include/domain_bridge/qos_options.hpp
ivanpauno: yes

ivanpauno: Fixed in https://github.com/ros2/domain_bridge/pull/73/commits/4e66a4fad7e145e9beae59831c1b66c9d3869ef7

adityapande-1995 (approved):
Tested locally, works !

ivanpauno:
Thanks for reviewing @adityapande-1995 !!!

PR #79, opened 2023-10-18
fix bidirectional bridge looping over the messages with fastdds

cyrilleberger:
When using fastdds (default in humble), and using a bidirectional topic bridge, the same message keep been sent over and over. This is because the behavior of `ignore_local_publications` is different between fastdds and cyclonedds, this is was reported to `rmw_fastrtps` in https://github.com/ros2/rmw_fastrtps/issues/573. To fix this problem, I made a change in domain_bridge to make sure that the message we receive from a subscriber do not come from a publisher owned by the bridge. This fix the problem of sending the same message over and over.

### DIFF
--- a/src/domain_bridge/domain_bridge.cpp
+++ b/src/domain_bridge/domain_bridge.cpp
@@ -49,6 +49,87 @@
 
 namespace domain_bridge
 {
+/// \internal
+/**
+ * A hack, because GenericSubscription doesn't allow to define callback that takes the
+ * serialized message and the message type info at the same time. See issue #1604 of rclcpp.
+ */
+class GenericSubscription : public rclcpp::GenericSubscription
+{
+public:
+  template<typename AllocatorT = std::allocator<void>>
+  GenericSubscription(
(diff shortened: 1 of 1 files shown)


src/domain_bridge/domain_bridge.cpp
```diff
@@ -234,6 +320,17 @@ class DomainBridgeImpl
     return publisher;
   }
 
+  /// Check if there is a pulisher with the given gid
+  bool has_publisher_with_gid(const rmw_gid_t & _publisher_gid)
+  {
+    for (auto it = bridged_topics_.begin(); it != bridged_topics_.end(); ++it) {
+      if (*it->second.first->get_base_publisher() == _publisher_gid) {
```
AdamGoertz, line 327: Thanks for looking into this, we were having the same issue and this PR helped us. I noticed one issue when we tried using this; we got a segmentation fault on this line. I haven't looked into the root cause, but adding this patch fixed the issue.

```suggestion
      if (it->second.first && *it->second.first->get_base_publisher() == _publisher_gid) {
```

PR #82, opened 2024-09-15
Adds support for runtime configuration of actions and services via code generation

taDachs:
This RP implements support for the runtime configuration of services and actions. This is accomplished by using code generation to generate template code for each action and service in the currently sourced workspace at compile time, similar to how the ros1_bridge handles it.

Looking forward to your feedback

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,8 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_python REQUIRED)
+
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_components REQUIRED)
 find_package(rcutils REQUIRED)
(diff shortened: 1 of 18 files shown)


taDachs:
Would it be possible to get some feedback on this PR? I realize that code generation might not be the most elegant solution, but from what I understand, runtime introspection for services is only available from Jazzy onwards, and action introspection isn’t yet supported.

If there’s a better or more idiomatic way to approach this that I might have missed, I’d really appreciate any suggestions and would be happy to update the implementation accordingly.

alireza-moayyedi:
I was having the exact same issue (though on Jazzy) and stumbled upon this PR. Nice work @taDachs !

@jacobperron would it be possible to move forward this PR (or any other solution in general to allow runtime service/action bridge configuration)?

ralilahlou-spec:
I tested the code several times on a domain_bridge setup on Humble, and I can confirm that this effectively works out of the box without any issues except for a moderately slower compilation time. Are there any plans for a merge on this ? The feature is implemented soundly, really useful and avoids needing to fork domain_bridge whenever services or actions need bridging.

Issue #85, opened 2025-01-06
rolling github workflow and Rpr job is failing.


fujitatomoya:
see the following failures.

- https://github.com/ros2/domain_bridge/actions/runs/12551337060/job/34995593133?pr=84

```bash
/usr/bin/bash -c source /opt/ros/rolling/setup.sh && colcon build --symlink-install --packages-up-to domain_bridge --event-handlers=console_cohesion+
  ERROR:colcon.colcon_core.entry_point:Exception loading extension 'colcon_core.environment_variable.powershell_command': cannot import name 'OVERRIDE_OPT' from 'em' (/usr/local/bin/em.py) The Python package 'empy' must be installed and 'em' must not be installed since both packages share the same namespace
  Traceback (most recent call last):
    File "/usr/local/lib/python3.10/dist-packages/colcon_core/entry_point.py", line 120, in load_entry_points
      extension_type = load_entry_point(entry_point)
    File "/usr/local/lib/python3.10/dist-packages/colcon_core/entry_point.py", line 168, in load_entry_point
      return entry_point.load()
    File "/usr/local/lib/python3.10/dist-packages/pkg_resources/__init__.py", line 2465, in load
      return self.resolve()
    File "/usr/local/lib/python3.10/dist-packages/pkg_resources/__init__.py", line 2471, in resolve
      module = __import__(self.module_name, fromlist=['__name__'], level=0)
    File "/usr/local/lib/python3.10/dist-packages/colcon_powershell/shell/powershell.py", line 17, in <module>
      from colcon_core.shell.template import expand_template
    File "/usr/local/lib/python3.10/dist-packages/colcon_core/shell/template/__init__.py", line 19, in <module>
      raise e from None
    File "/usr/local/lib/python3.10/dist-packages/colcon_core/shell/template/__init__.py", line 10, in <module>
      from em import OVERRIDE_OPT
  ImportError: cannot import name 'OVERRIDE_OPT' from 'em' (/usr/local/bin/em.py) The Python package 'empy' must be installed and 'em' must not be installed since both packages share the same namespace
``` 

- https://build.ros2.org/job/Rpr__domain_bridge__ubuntu_noble_amd64/1/

clalancette:
This almost certainly means that EmPy 4.X is being installed in the action.  The fix probably involves making sure to use the .deb package of empy (which should be 3.4.x).

fujitatomoya:
[...]

PR #86, opened 2025-01-09
fix github workflow for rolling.

fujitatomoya:
closes https://github.com/ros2/domain_bridge/issues/85

### DIFF
--- a/src/domain_bridge/component_manager.cpp
+++ b/src/domain_bridge/component_manager.cpp
@@ -22,7 +22,15 @@ namespace domain_bridge
 rclcpp::NodeOptions
 ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> request)
 {
-  auto options = rclcpp_components::ComponentManager::create_node_options(request);
+  rclcpp::NodeOptions options;
+
+  // TODO(@fujitatomoya): remove this try-catch block after ComponentManager::create_node_options
+  // is fixed to avoid throwing exceptions for unknown parameters in the base class.
+  //   https://github.com/ros2/rclcpp/pull/2723
+  try {
+    options = rclcpp_components::ComponentManager::create_node_options(request);
+  } catch(...) {
+  }
 
   for (const auto & a : request->extra_arguments) {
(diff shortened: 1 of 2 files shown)


christophebedard:
The empy issue mentioned in #85 is now gone, but the GitHub job now fails just like the Rpr job.

fujitatomoya:
@christophebedard thanks for letting me know, i will take a look!

fujitatomoya:
i confirmed with local environment that https://github.com/ros2/rclcpp/pull/2723 can fix the test failure happening in github workflow for rolling and rpr job.

src/domain_bridge/component_manager.cpp
```diff
@@ -22,7 +22,15 @@ namespace domain_bridge
 rclcpp::NodeOptions
 ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> request)
 {
-  auto options = rclcpp_components::ComponentManager::create_node_options(request);
+  rclcpp::NodeOptions options;
+
+  // TODO(@fujitatomoya): remove this try-catch block after ComponentManager::create_node_options
+  // is fixed to avoid throwing exceptions for unknown parameters in the base class.
+  //   https://github.com/ros2/rclcpp/pull/2723
```
fujitatomoya, line 29: i think it takes some time until https://github.com/ros2/rclcpp/pull/2723 is synced and released in rolling distro, so ended up having the temporary work-around to make domain_bridge workflow and rpr job green.

fujitatomoya:
@christophebedard all green :+1:

src/domain_bridge/component_manager.cpp
christophebedard: You wouldn't want to wait until the rclcpp PR is merged and released?

fujitatomoya: i think it could be merged soon, but releasing can take time. besides i do not think this work-around does not bring any downsides. or we should wait it out?

christophebedard: I was just wondering if there was any reason _not_ to wait, but I don't mind.

christophebedard (approved):

PR #88, opened 2025-12-05
fix:Qos type mismatch

dreamer198:
## Description

Fixes # (issue)


mjcarroll:
Can you add some details here?  Which distribution were you seeing issues on that this fixes?  CI is only passing for Jazzy right now, broken on Kilted and Rolling.

PR #89, opened 2026-04-03
Don't use rosbag2_cpp/typesupport_helpers.hpp for Jazzy and later

wentasah:
## Description

This fixes the following compile error in Jazzy:

    src/domain_bridge/domain_bridge.cpp:38:10: fatal error:
    rosbag2_cpp/typesupport_helpers.hpp: No such file or directory
       38 | #include "rosbag2_cpp/typesupport_helpers.hpp"
          |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

It happens since rosbag2 release 0.26.10 and is caused by ros2/rosbag2#2017 and its backport to Jazzy (ros2/rclcpp#2902).

We keep the previous implementation, which is needed for Humble and add a new one, which works for Jazzy and later.

### Is this user-facing behavior change?

No

### Did you use Generative AI?

No

### DIFF
--- a/src/domain_bridge/domain_bridge.cpp
+++ b/src/domain_bridge/domain_bridge.cpp
@@ -35,7 +35,13 @@
 #include "rclcpp/generic_subscription.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp/serialization.hpp"
+#include "rclcpp/version.h"
+#if RCLCPP_VERSION_MAJOR >= 25
+#include "rclcpp/typesupport_helpers.hpp"
+#else
+// Humble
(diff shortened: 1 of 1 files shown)


christophebedard:
could you test on Rolling?

wentasah:
Unfortunately not - for two reasons. One is #90. If I try the version before `zstd_vendor` removal, I get failures due to `Unknown CMake command "ament_target_dependencies"`. As far as I know this is another thing that was changed in rolling and projects need to be updated to use plain CMake dependencies.

christophebedard:
ah yeah, this needs a bit of work to get working again on Rolling.

Since this seems correct, I'll just merge this PR.

